Issue #13, opened 2026-07-30
A working tree with a truncated .git/index aborts instead of reporting


breferrari:
Found while gating I2a, in the test that asserts a failed `advance()` leaves the previous frame intact.

## What happens

`gix` 0.86 (via `gix-index` 0.54) reads the index trailer as `data[data.len() - object_hash.len_in_bytes()..]` with no lower bound. An index file shorter than the hash underflows that subtraction, so opening the repository **panics** instead of returning an error.

Measured on Windows, debug build, by replacing `.git/index` with N bytes and calling the change walk:

| index size | outcome |
|---|---|
| 0 bytes | panic (`attempt to subtract with overflow`) |
| 1 byte | panic |
| 19 bytes | panic |
| 20 bytes | `Err`: shared index checksum mismatch |
| 21 bytes | `Err` |
| 40 bytes | `Err` |

So the boundary is exactly the SHA-1 length. Every other malformed shape tried reports cleanly: 100 bytes of garbage, a valid `DIRC` header with an impossible entry count, and an unsupported version number all return `Err`.

A zero-length index is the realistic case. It is what a crash or a full disk during an index write leaves behind, and it is what `> .git/index` produces.

## Why it matters here

vigia reads the index on every tick, for days. The release profile sets `panic = "abort"`, so this cannot be caught: the process dies with no message and, once there is a TUI, a terminal left in whatever state it was in. That is the failure mode I8 exists to forbid.

## Why it is not fixed in this pass

It is a defect in a dependency, not in the frame path, and the plausible local defences are worse than the problem. A pre-flight length check on `.git/index` guards this one shape while implying the decoder is otherwise safe, which is a claim nothing here has established. `catch_unwind` does not apply under `panic = "abort"`.

What #2 does gate is the property vigia owns: given an `Err`, the previous frame survives it. See `a_failed_advance_leaves_the_frame_intact`, which uses 128 bytes of garbage deliberately, with the boundary documented at the call site.

## What to do

1. Report upstream against `gix-index`. The fix is a length check before the slice.
2. Decide vigia`s own posture once I8 lands, since that is where terminal restoration on panic is settled anyway.

Filed rather than absorbed, per the deferral rule in `ROADMAP.md`.

PR #14, opened 2026-07-30
I2a: the frame path never re-diffs what did not change

### LINKED ISSUE #2: I2a: the frame path never re-diffs what did not change
The frame path must never re-diff a file that did not change.

**Budget:** re-diff cost proportional to what changed, not to the size of the working tree.

**Why it is load bearing rather than an optimisation.** Measured during the Phase 1 spike, release build, on a 100k-line diff: re-diffing every changed file costs **18.58ms p99**, over the 16ms I9 budget, against **3.27ms** for a single file. So the naive frame path breaks I9 on its own.

**Proof:** assert the re-diff count and byte count for a single-line edit in a large working tree.

## Note on numbering

I2 in `SPEC.md` originally read as "re-highlighting is incremental", which conflates two invariants with different dependencies. Incremental re-diffing needs only `gix` and belongs to Phase 1. Incremental re-highlighting needs `syntect` and is tracked separately as I2b.

breferrari:
Closes #2. Phase 1 closes with it.

## The problem

Re-diffing every changed file costs **18.58ms p99** on a 100k-line diff against a **16ms** frame budget. The naive frame path broke I9 on its own, so I2a is load bearing rather than an optimisation.

## What lands

`Frame` holds the previous frame's diff per path and revalidates instead of recomputing. Validity is decided from three things, none of which reads a file:

- the **index blob** the change names, because the index is the left-hand side of every diff drawn, and staging moves it without touching the disk
- the **kind** of change, because a rename carries the path it came from and a removal has no working-tree side
- a **`stat`** of the working-tree file: size and modification time

Content is never hashed to decide this. Hashing is the read I2a exists to avoid.

Other properties it carries:

- `advance()` re-reads the file list and **diffs nothing**, so a caller still paints the top of a long diff without touching the bottom (I4). It does walk status to completion, because a scrollbar needs the file count, and that costs nothing while rename tracking is on and cannot stream either. `SPEC.md` section 10 tracks it.
- `advance()` commits nothing until the walk succeeds, so a failed status walk leaves the previous frame intact rather than blanking the pane.
- Paths that stop being changed are dropped, keeping the held set proportional to the diff rather than to the session (I3).

## The rule that was wrong, and now is not

This is the part worth reading. The first version of the staleness rule trusted a fingerprint when its modification time was **earlier than the moment the content was read**, on the reasoning that any later write must move the time forward. That reasoning is false, and the audit caught it.

A filesystem stamps a modification time by **flooring** the current time to its own granularity. So a write landing *after* the read can record a time *before* it, leaving `(len, mtime)` bit-identical. A second same-length write inside that granule is then invisible, and stays invisible: the reuse path returns without refreshing the fingerprint, and I1 guarantees no timer will ever look again. The pane keeps showing a diff that does not exist.

Reproduced before changing anything, driving the public API against an oracle with no memory:

| | stale diffs | attempts | shared an mtime with their predecessor |
|---|---|---|---|
| before | **10** | 4000 | 10 |
| after | **0** | 4000 | 12 |

The stale count matched the shared-mtime count exactly, 10 of 10, which pins the mechanism rather than merely correlating with it. NTFS granularity on this machine is about 1ms. On a 1s volume (HFS+, ext3) it would be most rewrites rather than one in four hundred.

The rule now requires a **full granule** to have passed between the stamp and the read. The margin has to be an upper bound on real granularity, so 2 seconds: FAT and exFAT quantise to 2s, HFS+ to 1s, NTFS and ext4 and APFS to milliseconds or finer. It costs redundant diffs of files written in the last two seconds, which are files that just changed.

Git avoids this differently, by comparing against the index file's own modification time so both sides come from the same clock at the same granularity. That option is not available here: every file the frame path cares about is dirty against the index, so every one of them would be permanently suspect and I2a would be dead. Hence the margin.

**The deeper miss was in the tests, and it is the more useful lesson.** Four mutations of the rule had already been killed, and all four were in `reusable`, which *consumes* the decision. Nothing tested the code that *produces* it, so hardcoding the answer to "trusted" passed the entire suite. Both halves are covered now.

## Numbers

Release build, 100 files, 100k-line diff, every file materialised:

| | p99 | bytes read |
|---|---|---|
| cold frame, nothing to reuse | 18.28ms | 3.6 MiB |
| revalidating frame, nothing edited | 3.93ms | **0 B** |
| **real frame, one line rewritten before each frame** | **6.97ms** | one file |

The last row is the gated one, and it is what I9 now measures: 250 sampled frames after a 50-frame warmup, 250 recomputes and 24750 reuses, inside the 16ms budget. The cold frame is the only shape that breaches it.

## How it is gated, in both directions

The two failure modes are not symmetrical, so they get different gates. **Reusing too little** is slow and loud. **Reusing too much is fast, passes every budget, and shows a diff that no longer exists**, which is the one worth building tests around. That asymmetry is now a rule in `SPEC.md` section 7 rather than a habit.

- `tests/budgets.rs` pins the recompute count at exactly **1** and the byte count **identical** across a 4-file and a 100-file fixture. Two fixtures rather than one call against the sum of all calls, because that form passes against the very regression it is written to catch: when every call inflates by N the sum inflates by N too and the ratio never moves.
- `tests/frame.rs` compares every reused frame against one computed with no memory at all, after a one-line edit, a same-length edit, staging, an index entry rewritten under an untouched file, a deletion, and a revert.
- `settled` is a pure function, tested over a table of real filesystem granularities from 100ns to 2s. It cannot be tested by racing a filesystem: the window is smaller than the machinery needed to hit it.
- `a_file_written_moments_ago_is_never_reused` covers the producer and needs no race to be deterministic. A recently written file must be re-read on *every* frame until its granule provably closes.

### Mutation tested

| mutation | caught by |
|---|---|
| never reuse | budget gate plus 9 behavioural tests |
| trust the index side alone | 3 behavioural tests plus budget gate |
| drop the racily-clean guard | unit test only (the window cannot be raced) |
| drop the index blob check | unit test plus `a_new_index_blob_invalidates_...` |
| **producer hardcodes "trusted"** | `a_file_written_moments_ago_is_never_reused` |
| **margin removed from the rule** | same |
| margin set to zero, or to 1ns | granule table plus the producer test, independently |
| `settled` returns always true, or always false | granule table, boundary tests, and every settling test |
| `<=` relaxed to `<` at the boundary | `a_read_a_full_margin_after_the_stamp_is_settled` |
| the working-tree side never checked | the I9 gate's content assertion, plus 3 behavioural tests |

## Harden audit

No prior merged PR to set the bar (this repo has none), so the bar was the landed watch engine for I1: eight behavioural tests each carrying a non-vacuity guard, a three-OS matrix, and a documented reason on every constant.

**Round 1** ran three agents with disjoint remits (adversarial, correctness and API, docs and conventions).

Real bugs found and fixed here:

- **The staleness rule compared clocks of different resolution.** `frame.rs`. Described above. Independently reproduced before acting on it.
- **I9 was gated over the primitives, not the frame path.** `tests/budgets.rs`. The steady-state number this PR claims was gated nowhere: the old gate timed one enumeration plus one diff, a frame shape that stops existing once I2a is in. New gate added, and the number in the spec is now the measured one.
- **The producer of the staleness rule had no test at all.** Four mutations had been killed in its consumer, which read as coverage and was not.
- **`SPEC.md` section 3 prescribed the weaker form of the I2a proof**, the single-fixture one the implementation deliberately rejected.
- **README claimed incremental re-highlighting**, which is I2b and not built, while omitting the re-diffing that is. Phase 1 still showed in progress, `notify` was missing from the dependency table, and Status linked only the spec where `ROADMAP.md` is the live state.
- **`hunk.rs` and `benches/engine.rs` still said bare `I2`.** An earlier commit here claimed to have corrected those and had only done `main.rs`.
- Smaller: `probes` under-described its own arithmetic; `lib.rs` promised the iterator shape was load bearing then pointed at a type that collects; `advance()` cited I4 without the file-list caveat; section 10 framed the completed walk as an open risk while the harness asserted it as settled; two 18.x measurements of the same quantity read as a typo of each other; three commit messages had a backtick where an apostrophe belonged.

**Round 2** re-audited the fix and its blast radius specifically. It confirmed the new rule sound for a local worktree, including that computing `settled` once at read time is not merely acceptable but required (recomputing it later would reintroduce exactly the round-1 bug). It also found three things:

- **The new I9 gate's non-vacuity assertion had gone vacuous.** It counted recomputes to show the edits were landing, which stopped meaning anything once the margin existed: one write keeps its file unsettled for the whole margin, so every frame in that window recomputes it whether or not anything is still being edited. Demonstrated by editing once instead of once per frame, which still passed. Now it asserts the frame is handing back the text of the last edit made, which the margin cannot manufacture.
- **The overflow arm in `settled` was unreachable and so untestable.** Rewritten with `duration_since`, which turns a future modification time into the same "cannot prove" answer through a case that *is* reachable and covered.
- **The margin's cost is not only redundant work.** Measured: after a bulk rewrite of all 100 files, every frame recomputes everything for the full margin, 18 to 21ms per frame, putting 82 of 620 consecutive frames over the 16ms budget for about two seconds. Single-file editing never approaches it (503 redundant diffs over 2.004s, nowhere near budget) and autosave every 500ms stayed under on all 973 frames. Recorded in section 10 next to the narrowing that would remove it, since it is the strongest argument for doing that narrowing.

Two mutations survive, both deliberate and both logged: moving the read timestamp after the read widens the trusted window by one file read, which matters only if a single diff exceeds two seconds; and one budget test sleeps 2.5s twice where once would do, which would cost passing the worktree into the settle helper at ten call sites to fix.

One round-1 finding did **not** survive checking: that the harness's cold byte count was cumulative over twenty frames. The `Frame` is recreated inside the loop, so it is a single frame's reading and the columns are comparable.

Judgement calls, logged rather than silent:

- **`Frame::diff` still panics on an out-of-range index**, and that is deliberate. A monitor holding a selection across frames must clamp it to `files().len()` for its own correctness anyway, or it renders a stale selection. A lenient accessor would convert a loud programmer error into a silent wrong row, so slice semantics are the better contract. Documented at the call site.
- **No property test over the reuse rule.** Its state space is finite and tiny and the unit tests already cover every reachable class; a generator would re-derive the same truth table while encoding the rule it checks. Where one would earn its place is a stateful model over random edit sequences asserting the frame equals a fresh diff, which is a bigger piece of work than this issue.

One pre-existing flake was fixed rather than re-run: `a_burst_of_writes_becomes_one_tick` wrote thirty files into a directory it created with the first of them, so it was racing inotify's new-directory watch rather than testing coalescing. It failed on a CI run whose tree was byte-identical to a passing one, which is what said flake rather than regression. I1 itself was fine: every tick triggers a full status walk, so the lost events cost a monitor nothing and only the test's counter cared.

Deferred, each on the shelf in `ROADMAP.md` with its reason:

- #13 a `.git/index` shorter than the object hash makes `gix` abort instead of reporting
- #15 a symlink diffs as its target's contents rather than as the link, demonstrated against git as the oracle
- #16 the fingerprint cannot see a write that restores the modification time, which no margin fixes and which needs metadata `std` does not expose on Windows
- #17 two paths differing outside UTF-8 collapse onto one cache key
- #18 a whole file is read to discover it is binary
- #19 an idle frame is one `stat` per changed file, 36.71ms at 2000 of them

## Scope of checks

Code diff, not docs-only, so the full suite and the budget gates apply.

**Tests: 25 to 53.** 14 unit, 11 frame, 8 watch, 14 fidelity, 4 budgets, 2 doc, plus 1 ignored diagnostic. Green in debug and in release with the absolute gates asserting at **zero slack**, and repeated 12 times in debug and 8 in release because the logic is timing-sensitive and one green run is not evidence. Three OS matrix green. `clippy --all-targets -D warnings` and `fmt --check` clean.

No dependency, flag or subcommand added: `Cargo.toml` and `Cargo.lock` are untouched.


breferrari:
Follow-up queued behind this: #20 adds a spec-versus-tracker pre-flight to `take-next` step 1. Held until this merges so an unrelated docs change does not land in this PR.

Issue #81, opened 2026-08-01
A washed row may be reaching the scrollbar column, or the terminal is


breferrari:
Reported from a real pane while scrolling a large diff: the removed-line wash appears to reach the far right of the screen and meet the scrollbar rather than stopping short of it.

## What is not yet established

**This may not be a `vigia` defect at all**, and it is filed with that said rather than after a diagnosis that assumes it. `with_bar` draws the bar and then hands the caller a narrowed rect:

```rust
self.scrollbar(region, at, span, of);
Rect { width: region.width.saturating_sub(BAR_WIDTH as u16), ..region }
```

and `line_row` washes only its own `area`. If the row receives the narrowed rect, the columns the bar occupies are not washed and the report is explained by something else — most plausibly the **host terminal's own scrollbar** overlaying the full-width band, which sits in exactly that place and is a similar colour.

So the first task is to reproduce and separate those, not to change the drawing.

## Reproduce

A diff long enough for the bar to appear, scrolled into a run of removed lines, on a pane wide enough that `wide` is true. Then decide between:

- **The wash covers the bar's columns** — the narrowed rect is not reaching `line_row`, or the wash is applied before the narrowing.
- **The drawing is correct** and the overlap is the host terminal's scrollbar over the pane.

`TestBackend` settles it without a screenshot: assert the cell styles in the last `BAR_WIDTH` columns of a washed row. That gate can fail and does not currently exist. [#63](https://github.com/breferrari/vigia/issues/63) is the neighbouring case — the wash dropping a column under a wide glyph — and one test shape reaches both.

## Why it matters either way

If it is the drawing, a wash under the thumb makes the position indicator hard to find in exactly the situation it is most needed: a long file, deep in a change.

If it is the host, the design question is whether a full-bleed row wash is right when terminals commonly draw their own chrome over that column — a ruling worth writing down once rather than rediscovering per host.

## Exit criteria

- A gate asserting what the last `BAR_WIDTH` columns of a washed row contain.
- The cause named, and if it is the host, recorded in §11.1 so the next report closes in one read.

Issue #80, opened 2026-08-01
The hint bar names a key twice and never names the arrows


breferrari:
The hint bar gained a fourth item, and it spends its most expensive columns restating a key the reader has already learned while never naming the keys hardest to guess.

## What it draws now

```rust
const HINT_RUNGS: [&str; 5] = [
    "q quit · f follow · jk scroll · JK files",
```

Two observations, pulling in opposite directions on the same line:

**`jk scroll · JK files` spends eleven columns on a case distinction.** A reader who has learned `jk` needs less to see `JK` spelled out as a peer item than to know the *shifted* pair addresses the other region. As two sibling items they read as four unrelated keys rather than one rule with two halves.

**The arrows are bound and advertised nowhere.** `KeyCode::Down | KeyCode::Char('j')`, `KeyCode::Up | KeyCode::Char('k')`, and a modifier path over `Down`/`Up` — all real, none named. §11.1's own argument for keeping `f follow` at the narrowest rung is that it is *"the one nobody would guess"*. By that reasoning an arrow is the one everybody guesses and needs no advertising, which is a good argument — but it is nowhere written down, so the omission reads as an oversight rather than a ruling.

## The rule the ladder already has, applied one level in

§11.1 rules the hint bar as a list that drops **whole** items, `jk scroll` first, leaving `f follow` last, because state outlives advice and the unguessable outlives the reflexive. That ordering was settled for four items and there are now five.

What is unsettled is whether an item may express a *modifier relationship* rather than a key — one item covering both regions instead of two items covering one each. That is a spec question about what a hint is, not a copy edit, and it wants ruling before a sixth item arrives.

## Exit criteria

- The rungs are re-ruled for the two-region pane, with the drop order stated.
- Whether arrows are deliberately unadvertised is written into §11.1 either way.
- The sweep in `legibility.rs` covers whatever the new rungs are; it asserts against the constant rather than a snapshot, so it follows automatically once the constant is ruled.

breferrari:
**Re-housed from the Shelf to Phase 8 on 2026-08-15, and the reason is that this issue's own premise expired.** Raised again by a reader looking at the footer while a pass ran, which is what [#76](https://github.com/breferrari/vigia/issues/76) says has to happen to a deferral: a reason is a dated claim and nothing re-reads one.

[...]

Issue #79, opened 2026-08-01
The wheel ignores the pointer, and the thumb it draws cannot be grabbed


breferrari:
The pane now has two scrollable regions and one wheel that always drives the lower one, and it draws a scrollbar thumb that cannot be grabbed.

## The wheel ignores where the pointer is

`mouse_action` matches on `mouse.kind` alone:

```rust
fn mouse_action(mouse: &MouseEvent) -> Option<Action> {
    match mouse.kind {
        MouseEventKind::ScrollDown => Some(Action::Scroll(WHEEL_ROWS)),
        MouseEventKind::ScrollUp   => Some(Action::Scroll(-WHEEL_ROWS)),
```

`MouseEvent` carries `column` and `row`, and neither is read. That was correct while the body was one region. With the file list a region of its own, the wheel over the list scrolls the diff underneath it — the one behaviour a pointer device is expected never to do.

`Action::ScrollList` already exists and is bound to `J`/`K`, so routing is the whole change: decide the region from `mouse.row` against the same layout the renderer used, and emit `Scroll` or `ScrollList` accordingly.

## The thumb is drawn and inert

There is no `MouseButton`, `Drag` or `Down(` anywhere in `crates/vigia/src/`. The existing comment says clicks and drags do nothing "because nothing is selectable in a monitor", which was true when written and is not the same claim as *nothing is draggable*.

A scrollbar with a thumb is a **published affordance**. Drawing one that does not respond to being grabbed is the aspirational-UX trap: the picture makes a promise the binary declines, the same class as [#66](https://github.com/breferrari/vigia/issues/66) and [#67](https://github.com/breferrari/vigia/issues/67), arriving through an interaction instead of a drawing.

Two rulings, and the second could go wrong quietly:

- **Drag the thumb** — track a drag origin and map row delta to position delta in the bar's own units, which `scrollbar` already computes in one place for both callers.
- **Click the track** — page toward the click, or jump to it? Paging is the pager convention and cannot overshoot; jumping is what a map affords, and the list already reads as a map.

Neither is selection, so neither costs the reviewer-class boundary §2 draws.

## Exit criteria

- The wheel acts on the region under the pointer, gated at a row inside each region.
- The thumb is draggable, and track-clicking is ruled either way in §11.1.
- The comment in `input.rs` is corrected: what is refused is *selection*, not pointer interaction.

Issue #78, opened 2026-08-01
Every file has an empty sparkline at launch, which is the common case


breferrari:
The mockup draws a sparkline on all three files, including the cold one. On a real launch there are none, and that is by construction rather than by defect.

## What happens

`spark_of` returns `None` when the history is empty:

```rust
if peak == 0 || buckets.iter().all(|&count| count == 0) {
    return None;
}
```

and `Row::File::spark` already names who that is:

> All zeroes for a file `vigia` has not seen change, which is the ordinary case for a worktree that was already dirty at startup.

I10's store is fed from the **watch**, one coalesced tick at a time. A file already modified when `vigia` started has no ticks behind it, so no history, so no sparkline — and it never gets one until it changes again.

## Why that is the common case rather than an edge case

The pane is opened *because* work is in progress. A reader who starts `vigia` beside an agent that has already written twenty files sees twenty rows and zero sparklines, so the element §5 names as one of four differentiators is absent from the first screen. The picture shows the opposite, including a dimmed `Cargo.toml` — a settled file — carrying a sparkline of its own.

Reported as *"that graph thing was never realized"*, which is the honest reading from outside: the feature is built, gated and correct, and invisible when it matters most.

## What it is not

**Not a case for backfilling from git.** Churn over the last 120 seconds is not recoverable from commit history, and inventing it would put a fabricated number in the one element whose whole job is to be trusted at a glance.

The real options are rulings rather than fixes:

- **Draw an empty ramp** rather than nothing, so the column exists and reads as "no history yet" instead of as an absent element. Costs the column at every width.
- **Say it once**, in the empty state or the header: history begins at launch.
- **Accept it and correct the picture**, since `assets/preview.svg` is a specification and currently shows a state the tool cannot produce on its first frame.

The third is cheapest and least honest to the design, which is why this wants a decision rather than a patch.

## Exit criteria

- A ruling in §5.1 for what a file with no history draws.
- If the answer is "nothing", the mockup is corrected so it stops promising otherwise.
- A gate over the launch-into-a-dirty-worktree case, which no fixture covers today: every history test drives ticks first.

breferrari:
## Plan

Ruled towards the first of the three options, in the form that makes it one rule rather than a special case: **a bucket with no writes draws a track, and a file with no history is the all-empty case.** Track glyph `_`.

### What the record gave, including what argues against this

[...]

PR #82, opened 2026-08-01
Re-house Phase 4, and the five findings that exposed it

breferrari:
Tracker rows and a shelf entry for five findings from running [#71](https://github.com/breferrari/vigia/pull/71)'s branch. **Documentation only — no code, no tests.**

Filed as [#77](https://github.com/breferrari/vigia/issues/77), [#78](https://github.com/breferrari/vigia/issues/78), [#79](https://github.com/breferrari/vigia/issues/79), [#80](https://github.com/breferrari/vigia/issues/80), [#81](https://github.com/breferrari/vigia/issues/81), each self-contained with its evidence so a fresh session can take any one of them cold.

## What was found

Two are **claims problems**, and they belong with [#66](https://github.com/breferrari/vigia/issues/66)'s siblings in Phase 4 because they are the same shape: the picture says something the binary does not do.

- **[#77](https://github.com/breferrari/vigia/issues/77)** — `assets/preview.svg` puts every glance element at a fixed x on every row (sparkline at 300, counts right-anchored at 452 and 500, heat from 540), so the columns align down the list and about 18% of the width is deliberate trailing space. `Painter::file_row` right-packs instead, so each element's position depends on the widths of the elements outside it, and `+139 -131` against `+293 -0` shifts a row three columns from its neighbour. The small-multiples reading the list exists for is gone, and the row reads as crammed because right-packing removes exactly the whitespace the mockup keeps.
- **[#78](https://github.com/breferrari/vigia/issues/78)** — every file has an **empty** sparkline at launch. `spark_of` returns `None` on an all-zero history, and I10's store is fed from the watch, so a worktree that was already dirty when `vigia` started has no ticks behind it. That is the ordinary case: the pane is opened *because* work is in progress. One of §5's four named differentiators is absent from the first screen, and the mockup shows the opposite on all three rows including the settled one.

Three are **interaction**, and they arrive with the second region rather than being pre-existing:

- **[#79](https://github.com/breferrari/vigia/issues/79)** — `mouse_action` matches on `mouse.kind` alone and never reads `column`/`row`, so the wheel over the file list scrolls the diff underneath it. And the thumb is drawn but inert: no `MouseButton`, `Drag` or `Down(` anywhere in `crates/vigia/src/`. A thumb is a published affordance, so drawing one that cannot be grabbed is the same aspirational-UX class as [#66](https://github.com/breferrari/vigia/issues/66) and [#67](https://github.com/breferrari/vigia/issues/67), arriving through an interaction instead of a drawing.
- **[#80](https://github.com/breferrari/vigia/issues/80)** — the hint bar spends eleven columns on `jk scroll · JK files` while never naming the arrows it binds. Whether that omission is a ruling or an oversight is not written down either way.
- **[#81](https://github.com/breferrari/vigia/issues/81)** — a washed row appears to reach the scrollbar column. Filed **undiagnosed on purpose**: `with_bar` narrows the rect before the rows draw, so this may be the host terminal's own scrollbar over a correct full-bleed band. The first task is a `TestBackend` gate over the last `BAR_WIDTH` columns, which can fail and does not exist.

## Why this is not on top of #71

Asked for as a branch stacked on `66-the-file-list-is-a-region`. It is cut from `main` instead, and the reason is mechanical: that branch is six commits behind `main` and predates the `#72`–`#76` rows, so adding `#77`–`#81` to its copy of the table would have produced a guaranteed conflict in the same rows on merge. Nothing here touches the code that PR changes, so stacking bought nothing that ordering does not.

**It does not block [#71](https://github.com/breferrari/vigia/pull/71) and [#71](https://github.com/breferrari/vigia/pull/71) does not block it.** Land them in either order.

## Why none of it is implemented here

[#71](https://github.com/breferrari/vigia/pull/71) is under `/harden` as this is written, and every one of these findings touches `render.rs` or `input.rs` — the files that pass is rewriting. Implementing against a moving base would collide with the audit rather than help it. The issues carry the full diagnosis so the work can start cold once the branch settles.


breferrari:
**Scope grew, and the second commit is the more important one.**

Filing the five findings pushed Phase 4 to eight open issues, of which exactly **one** — [#12](https://github.com/breferrari/vigia/issues/12) — distributed anything. Every other one had been filed there on the argument that it must land *before* shipping. That is a statement about **ordering**, and ordering had been quietly treated as **membership**.

Principle 1 at the top of `ROADMAP.md` says a phase name is a filter you can quote back at a proposal to kill or delay it. "Distribution" cannot reject a sparkline bug, so by the file'"'"'s own test the name had stopped working — and a milestone that accepts anything is one `take-next` cannot take from deliberately.

Split by what the work **is**:

| Phase | Filter | Holds |
|---|---|---|
| 4 — the artifacts tell the truth | does this close a gap between what a published artifact claims and what the binary does? | #66, #67, #77, #78 |
| 6 — measured, not assumed | can this be decided without a number? | #47, #72, #50 |
| 7 — distribution | does this put the binary in someone else'"'"'s hands? | #12 |

Phase 6 is separate from Phase 4 rather than folded into it because those three are blocked on **calendar** time, not engineering time — #72 needs a real working session and #47 needs a runner without a six-hour cap. Bundled together, a phase of one-afternoon changes would sit open behind a day of waiting, and a phase that cannot close stops being a unit of work.

Phase 5 keeps its number and moves to the end of the file. It is a shelf rather than a step, and a dozen shelf entries plus #19, #34 and #51 cite "Phase 5" by name, so renumbering to tidy the file would silently repoint every one of them.

**And splitting the phases broke a check nobody was watching**, now [#83](https://github.com/breferrari/vigia/issues/83): step 1 sorts milestones by `due_on`, every milestone here has none, so the order is undefined. It answered correctly by luck while two were open and the lower number was the one to take. Now number order and execution order have come apart and the one milestone that must **never** be selected — the shelf, 22 open — sits numerically between two that must. Silent, and in the first command of the pass.

Still documentation only. No code, no tests.

breferrari:
Superseded by #71, which merged this branch in rather than landing beside it. The ROADMAP rows and the shelf entry filed here are on `main` now, carried through #71's merge commit, and the re-housing they describe (Phase 4 renamed, Phase 6 and Phase 7 split out) is live.

Two of the five findings closed on #71's own branch rather than waiting for a phase: #79, because the pointer and the thumb are affordances that branch published, and #81, which asked for a gate rather than a fix and came back green. #77, #78 and #80 remain open.

Closing as merged-by-proxy rather than as rejected.

PR #116, opened 2026-08-05
process: the leak rule gets its hook, and the capture step gets a fallback

breferrari:
## What

House rule #1 (no agent-session artifacts in anything that lands in this repo) finally gets enforcement that lives where the risk does: a `PreToolUse` hook that inspects every `gh pr|issue|release create|edit|comment` and blocks the call when the body carries a session URL, a `Claude-Session:` trailer, a local absolute path, or a body flattened to a single line.

Also, `take-next` step 8 gains the fallback the capture step was missing: when the vault's `record_work` refuses a legitimate record, file the note by hand and say so in the report — and after any success, read the note back, because one record returned success while writing its sections as raw markup.

## Why now

The rule existed only as prose here, while the sole enforcing hook lived in a different working directory — so a session working from this checkout was unguarded exactly where the repo is public. `take-next` even asserted "that direction is the one with a hook guarding it", which was not true from this cwd; the sentence is rewritten to say precisely what is and is not covered (commit messages remain unguarded — that discipline stays manual).

## Verification

- Guard red-tested on all four artifact classes (session URL, trailer, absolute path, flattened body): each blocks with exit 2 and a named reason.
- Clean body, non-`gh` commands, and unparseable stdin all pass through (fail-open by design — a guard must never block work by breaking).
- `sh .claude/skills/take-next/selftest.sh`: all 18 checks pass, including the SKILL.md drift assertions.

Docs + tooling only; no Rust code touched, so the suite and budget gates did not run (the repo's own docs-diff carve-out, logged here per step 5).

### DIFF
--- a/.claude/scripts/leak-guard.mjs
+++ b/.claude/scripts/leak-guard.mjs
@@ -1,0 +1,70 @@
+#!/usr/bin/env node
+// PreToolUse guard: block a `gh pr|issue|release create|edit|comment` whose body
+// carries an agent-session artifact, before it reaches a PUBLIC repo. The rule
+// this enforces already lived in CLAUDE.md; this is the part that does not
+// depend on a session remembering it. Exit 2 blocks the call; anything the
+// guard cannot parse or read fails OPEN (exit 0) — a guard must never block
+// work by breaking.
+//
+// Classes checked (each is a class of artifact, not one incident):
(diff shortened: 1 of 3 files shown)


breferrari:
Landed on main as 7041ade (squashed).

Issue #149, opened 2026-08-14
decision: should the pinned list draw the rank the digits address?


breferrari:
## decision: should the pinned list draw the rank the digits address?

Found while shipping [#122](https://github.com/breferrari/vigia/issues/122), and filed rather than folded in because the answer is a layout change and #122 is not.

#122 left the digits `1`-`6` unhinted, and one of the two reasons it gave was that "like drag-the-scrollbar, a drawn numbered list *is* the affordance". **The list draws no numbers.** `render__the_glance_elements_at_eighty_columns.snap` is the evidence:

```
"  vigia · 3 changed                                                   watching  "
"  M src/engine/watch.rs                    ● ████████████ __▁▂▆▄▆█   +42    -7  "
"  M src/render/frame.rs                      ████████████ ___▂▁___   +11    -3  "
"  M Cargo.toml                                            ________    +2    -0  "
"  q quit · f follow · jk scroll · JK files                                 1/3  "
```

A caret column two wide, then the status letter, then the path. No rank anywhere.

The refusal to hint survived the correction on `SPEC.md` §11.1's other reason, which is #121's measured one: the eighty-column footer is exactly full, so a fifth hint is paid for out of the frame-time and memory cells. That is settled and this issue does not reopen it. What is *not* settled is the half the wrong premise was hiding: **the digits are the first binding in this shell with no drawn affordance at all.** The thumb is drawn and grabbable, the list is drawn and clickable, `f` and `q` are hinted. `1`-`6` are none of those, so today they are reachable only by someone who read the spec or the README.

### The cheap option, which is why this is worth a decision rather than a note

The caret column is already **two columns wide at every width** (`CARET_WIDTH` in `render.rs`), and it is blank on every row except the one the diff is inside. So the rank can be drawn for free:

```
1 M src/engine/watch.rs      ->   ▸ M src/engine/watch.rs   (current file keeps the caret)
2 M src/render/frame.rs
3 M Cargo.toml
```

No column is spent, no ladder rung moves, and I6 is untouched at forty columns. It is a **look** change rather than a layout one, which is exactly the Phase 8 class.

### What has to be weighed

- **It costs the glance.** §5 makes the four glance elements shape-and-colour precisely so the eye takes them without reading, and §11.1 adds that "text on screen is either content, a name, or advice". A rank is a fourth thing and it is *read*, in the leftmost column, ahead of the status letter and the path.
[...]

Issue #121, opened 2026-08-05
Half-page scroll: d and u, the less bindings, whole or not at all


breferrari:
## Half-page scroll is the missing pager reflex

The key table covers row scroll (`j`/`k`), whole-page (`Space`/`PgDn`/`PgUp`), and the ends (`g`/`G`) — but not the half-page, and the half-page is how `less` and vim users actually read code: whole-page loses visual continuity across the jump, row-scroll is too slow to traverse a hunk, half-page keeps an anchor on screen.

**The bindings are `d` and `u`, and they are `less`'s own.** `less` binds `d` to half-page-down and `u` to half-page-up, and this shell already claims `less +F` semantics for follow mode, so the precedent is internal as well as cultural. Both letters are free.

**`Ctrl-D`/`Ctrl-U` are explicitly not proposed**: `Ctrl-D` quits (it is in the quit row of the key table), and rebinding a quit key to a scroll is the kind of surprise the keymap has avoided so far. `d`/`u` alone, whole or not at all.

Mechanics inherit everything the existing scroll already settled: a manual scroll disengages follow, position stays `(file, offset)`, the past-the-end resolution applies. The hint bar does not grow — `d u` would ride as a bonus rung at widths with room, like `JK files`, and is never worth a footer row.

breferrari:
## Plan

Taken with `take-next`. Phase 8's topmost unstarted row.

### What the record gave me

`recall` returned *"Removing a rung from the top of a monotone layout ladder is free; from anywhere else it is not"* (verified 2026-08-03). The mirror case is the one here: adding a rung at the *top* of a written-out ladder moves no boundary beneath it, because nothing below refers to it. Confirmed against `render.rs::fitting`, which is a table walk (`find` the first that fits) rather than a greedy allocator. Its corollary, *predict then confirm*, is why this plan states the expected snapshot diff before anything runs.

Also stood on: `SPEC.md` §11.1's hint-bar ruling (a list drops whole hints, drop order `JK files` then `jk scroll` then `q quit`, `f follow` last standing), §11.1's bonus-rung rule and `HINT_BASELINE`, §11.1's drop order for the status readouts, and §5 line 201's rule that the mockup is corrected in the same pass as any ruling that changes what it draws. That last one has a precedent rather than an assumption behind it: `02ec750`, the pass that added the `JK files` bonus rung, edited `assets/preview.svg` in the same commit.

[...]

Issue #147, opened 2026-08-14
decision: a bonus hint rung is never worth a footer row, and nobody has ruled whether it is worth a readout


breferrari:
## The drop order puts advice above instrumentation, and a bonus rung is not advice anybody is owed

`SPEC.md` §11.1 rules the footer's drop order plainly: the hints are how a reader operates the tool and the state is what the tree is doing, so a narrowing pane owes them last, and *"a ladder that dropped a hint to keep a frame time would be legal by every other gate in this file, and wrong."*

`HINT_BASELINE` then carved out a second category that ruling predates. Rungs at or below the baseline are what a reader is **owed** at forty columns. Rungs above it are **bonus**: drawn where there happens to be room, and, per that constant's own reason, *"never worth a row."*

The question nobody has answered is whether a bonus rung is also never worth a **readout**.

### Why it is live rather than tidy

`Footer::plan` hands the diagnostics whatever survives `taken + CELL_GAP + hints`, so every column the hint bar takes is a column the readouts lose. Measured on `main` in the state that has both cells to draw (follow engaged, a position, `diagnostics()` chrome):

| Bar drawn | Frame cell arrives | Both cells arrive |
|---|---|---|
| today, widest rung 40 columns | 69 columns | 77 columns |
| with a 50-column rung above it | 81 columns | 89 columns |

The existing `JK files` bonus rung is already inside this. It is 11 columns wide, and dropping it would move both boundaries inward by about that much, which means the memory cell is being spent on a hint across roughly thirty widths right now. Nobody decided that; the allocation order decided it.

### Why [#121](https://github.com/breferrari/vigia/issues/121) did not settle it

That pass proposed `du half` as a bonus rung and measured what it cost: at eighty columns the footer row is 40 of hints, 2, 19 of readouts, 2 and 13 of state, which is exactly the 76 columns eighty leaves after the inset. A fifth hint fits only by taking both cells, and `render__the_status_bar_carries_what_a_frame_cost.snap` is the snapshot whose stated job is to show the layout is *good* at one width rather than legal at every width.

So the hint was dropped and `d`/`u` shipped unhinted, which is consistent with the unhinted `Space`/`PgDn`/`PgUp` beside them. That is a decision about **one hint**, taken conservatively because it needed no new ruling. It leaves the general question exactly where it was, and the next hint anybody wants will arrive at it again.

[...]

Issue #66, opened 2026-08-01
The mockup pins a file list above the diff, and the shell draws one stream


breferrari:
`SPEC.md` §11.2 **B4** is the last mockup-facing question still marked `(proposed)`, and it is phrased as a navigation question:

> **B4 — Is the file list navigable?** The README mockup shows a file list above the diff. Selectable, with the diff jumping to the selection, or a map rather than a menu?
>
> *(proposed)* **Not navigable in v1** — one continuous scroll, list as map.

Navigation is not the load-bearing half. **The picture and the shell disagree about whether the file list is a region of the screen at all**, and the proposal settles that silently, under the cover of answering something else.

## What the picture commits to

In `assets/preview.svg`, `src/engine/watch.rs` **is drawn twice**: once in the block of three summary rows at the top, and once as the diff heading below the rule. A single scrolling stream never draws one file's heading twice. The picture is a two-region layout — a summary list of every changed file, pinned above a rule, with one file's diff beneath it.

That is also where the picture's glance claim lives. §5 names four elements as the differentiator — sparkline, heat strip, `+`/`−` counters, pulse. Three of the four are drawn in that pinned block, **for all three files at once**, and they stay there while you read.

## What the shell does

One stream. `Row` is a flat enum in `crates/vigia/src/view.rs`:

```rust
pub enum Row { File { .. }, Hunk { .. }, Line { .. }, Note(..) }
```

`View::rows` is one `Vec<Row>`, `Position` is `(file, row)`, and `Painter::body` in `crates/vigia/src/render.rs` is a single loop over it. The sparkline, the counters and the heat strip are drawn on `Row::File` — a heading **inside** the scroll stream.

Nothing here is a defect. Every element is specified in §5.1, built, and gated. The container they were drawn for is what has no ruling behind it.

## Why it matters

**The glance signals are visible only while their heading is on screen.** Scroll two hundred lines into a file and every other file's heat strip, sparkline and recency dimming is off-screen. The glance surface disappears exactly while the pane is being read, which is the inverse of what the picture promises.

For a monitor this is the interesting direction of the loss. §2 buys "readable at a glance" as one of three legs; if the answer is the stream, then "at a glance" means *at the moment the diff is scrolled to the top*, and that is worth saying out loud rather than discovering from a screenshot.

[...]

PR #150, opened 2026-08-14
n and p step a file, the digits name a drawn row (#122)

### LINKED ISSUE #122: File-granular navigation: n/p next and previous, digits 1-6 jump to a listed file
## File-granular navigation: `n`/`p`, and the digits as direct jumps

The keymap has row granularity (`j`/`k`), page granularity (`Space`/`PgUp`/`PgDn`), and the two ends (`g`/`G`). It has no **file** granularity — and the file is the unit the list draws, the header counts, and the reader thinks in. Today moving to the next file means holding `j` through the remainder of the current one or reaching for the mouse.

Two additions, both mode-free:

1. **`n` / `p` — next / previous file.** The letters are free (no search exists to claim `n`), the mnemonic is bare, and `less` readers already carry the next-file reflex from `:n`/`:p`. Jump lands at the file's heading, same resolution rule as a list click; a manual jump disengages follow exactly as any manual scroll does.
2. **`1`–`6` — jump to the Nth listed file.** The list caps at six rows, so the digits map one-to-one to what is on screen — the numbered-jump grammar terminal monitors already taught their readers. A digit with no row (three files, key `5`) does nothing, loudly nothing — no bell, no error, the same as clicking empty list space. Digits address the *visible window* of the list, not absolute file indices: what you can see is what you can name.

## What this deliberately is not

- Not a selection or focus mechanism — the jump moves the diff, exactly as a list click does; B4's no-selection ruling is untouched.
- Not `Tab` cycling: `Tab` invites focus semantics, which invites modes.
- Not bound on `Ctrl` anything — the plain letters are free, and the one Ctrl precedent (`Ctrl-C`/`Ctrl-D`) means quit.

## Hint bar

`n p files` could ride as a bonus rung; the digits are deliberately unhinted — like drag-the-scrollbar, a drawn numbered list *is* the affordance, and six more columns of hint would cost the bar its one-line height at widths that matter.

## Gates

Key handling is table-driven; the additions extend the existing scripted-sequence snapshots (jump lands at heading; follow disengages; digit past the count is a no-op). No layout change, no budget surface.

breferrari:
Closes #122.

## What this is

The keymap had rows (`j`/`k`), screens (`Space`, `d`, `u`) and the two ends (`g`/`G`), and not the unit the pinned list draws, the header counts and the reader thinks in. Reaching the next file meant holding `j` through the rest of this one or reaching for the mouse, which already had the gesture.

`n` and `p` step one changed file. `1` to `6` put the diff at the Nth row the list is drawing. All of them land on the file's heading, which is what a list click and a follow jump already resolve to, and none costs a diff: a heading is row zero, so nothing asks how tall anything is and I4 never sees it. B4 stands, because all three move the diff and remember nothing.

## The picture

Twelve changed files, eighty columns, twenty-four rows. The caret and the `N/M` readout are what move.

```
=== opening screen, follow engaged ===
  vigia · 12 changed                                                  watching
  ▸ M src/mod_0.rs                           ████████████ ________    +6    -6 █
    M src/mod_1.rs                           ████████████ ________    +6    -6 █
    M src/mod_10.rs                          ████████████ ________    +6    -6 █
    M src/mod_11.rs                          ████████████ ________    +6    -6 ▕
    M src/mod_2.rs                           ████████████ ________    +6    -6 ▕
    M src/mod_3.rs                           ████████████ ________    +6    -6 ▕
  M src/mod_0.rs                             ████████████ ________    +6    -6 █
  @@ -1,6 +1,6 @@                                                              ▕
  1 -fn before_1() { let value = 7; }                                          ▕
  ...
  q quit · f follow · jk scroll · JK files                      follow ▶  1/12
```

```
=== after `n`: the next file, at its heading ===
  vigia · 12 changed                                                  watching
    M src/mod_0.rs                           ████████████ ________    +6    -6 █
  ▸ M src/mod_1.rs                           ████████████ ________    +6    -6 █
    M src/mod_10.rs                          ████████████ ________    +6    -6 █
    M src/mod_11.rs                          ████████████ ________    +6    -6 ▕
    M src/mod_2.rs                           ████████████ ________    +6    -6 ▕
    M src/mod_3.rs                           ████████████ ________    +6    -6 ▕
  M src/mod_1.rs                             ████████████ ________    +6    -6 ▕
  @@ -1,6 +1,6 @@                                                              █
  ...
  q quit · f follow · jk scroll · JK files                                2/12
```

Note the footer: `follow ▶` is gone. A file step moves the diff, so it disengages follow like any other jump.

```
=== after `6`: the sixth row the list draws ===
  vigia · 12 changed                                                  watching
    M src/mod_0.rs                           ████████████ ________    +6    -6 █
    M src/mod_1.rs                           ████████████ ________    +6    -6 █
    M src/mod_10.rs                          ████████████ ________    +6    -6 █
    M src/mod_11.rs                          ████████████ ________    +6    -6 ▕
    M src/mod_2.rs                           ████████████ ________    +6    -6 ▕
  ▸ M src/mod_3.rs                           ████████████ ________    +6    -6 ▕
  M src/mod_3.rs                             ████████████ ________    +6    -6 ▕
  ...
  q quit · f follow · jk scroll · JK files                                6/12
```

`6` names the sixth row, and the sixth row is `src/mod_3.rs`, because status order is lexicographic and `mod_10` precedes `mod_2`. That is the point of addressing the window rather than the changed set: what you can see is what you can name.

Generated by a throwaway diagnostic against `TestBackend`, in the shape of `list.rs::the_region_at_fifty_files`. It is not in the diff: it gates nothing, and the repository already carries one such picture on purpose.

## Three rulings the issue left implicit

Each has a gate whose inversion was mutated and killed.

**`p` from inside a file goes to the previous file, never to the top of the current one.** The pager reflex of "this section first" would make one key mean two things depending on where the viewport happened to be, which is the rule this whole map is built on, and `g` already reaches a top.

**Both ends refuse rather than clamp**, so neither key ever moves the view in the direction opposite to itself: clamping the index would give `n` at the last file a backwards jump to that file's heading. Follow is still disengaged there, where nothing moved, because on this map follow yields to intent. `g` at the top and `G` at the last file already behave that way, and `J` at the last window takes the opposite rule for a reason that does not apply here: the state it protects has nothing on screen to announce it, where `follow ▶` leaving the footer is visible.

**The digits address the drawn window**, which needs a second bound the click path never did. A pointer is bounded by the region and a keyboard is not, so a nine-row pane gives the list four rows while `5` and `6` stay under the reader's hand. Without the bound the digit resolves against the file list and the diff jumps to a file the reader cannot see they named. (The *other* bound, the one on the file count, turned out to be for something else entirely, and the section at the bottom of this body is that finding.)

## A finding: the issue's reason for not hinting the digits was wrong

#122 argued the digits need no hint because "like drag-the-scrollbar, a drawn numbered list *is* the affordance". **The list draws no numbers.** `render__the_glance_elements_at_eighty_columns.snap` is a caret column and then `M src/engine/watch.rs`, with no rank on any row, and the pictures above say the same thing. So the digits are the first binding in this shell with no drawn affordance at all.

The conclusion survives on #121's measured reason instead, which is stronger: the eighty-column footer is exactly full, so a fifth hint is paid for out of the frame-time and memory cells whichever hint it is. `SPEC.md` §11.1 now records the correction rather than repeating the analogy.

What the correction leaves open is filed as **#149**, a `decision` row in this phase: should the list draw the rank it is now addressable by? The caret column is already two columns wide and blank on every row but one, so the rank costs no columns at all, which is why it is worth a ruling rather than a shrug. Not folded in here: it is a look change, `assets/preview.svg` currently rules against it, and #122 promised no layout change.

## Scope

Code diff, so the full gate ran, not the docs carve-out. `Cargo.toml`, `Cargo.lock` and the workflows are untouched.

Nothing drawn changes, so no snapshot moved. The frame path, the watch and the diff oracle are untouched: this is `input.rs` and one `App::apply` arm.

## Gates

These are the numbers as the first commit stood. The final counts are at the bottom, after the `/simplify` pass added a gate.

- `cargo test --workspace`: **572 passed, 0 failed**.
- `cargo test --release --test budgets` and the release core gates: **22 passed, 0 failed**. I9 is a 16ms p99 and the warm frames come in at **1.314ms scrolling down**, 1.1792ms scrolling back, 181.9µs scrolling up. The bulk-rewrite frame inside the settle margin is p99 **16.7983ms**, which is where it has been. Nothing here is on the frame path, and the numbers say so.
- `cargo clippy --all-targets`: clean.
- `cargo fmt --all -- --check`: clean.
- **Nine mutations, all killed.** Each asserted an exact occurrence count before it applied, because a mutation that silently does not apply reads as SURVIVED and is indistinguishable from a blind gate: the digit's window bound deleted, the file step clamping instead of refusing, `p` stopping at the current heading, `File` not disengaging follow, the digit range one short of the cap, the digits counting from zero, `n` and `p` swapped, `0` bound, and `File` claiming it needs the height.

New gates, all of which fail against `origin/main`:

| Gate | What it pins |
|---|---|
| `input::n_and_p_are_the_file_step` | the bindings, and `N`/`P` staying inert |
| `input::every_row_the_list_can_draw_has_a_digit` | that the restated `'1'..='6'` cannot drift from `LIST_ROWS` |
| `scroll::n_and_p_step_one_file_and_land_on_its_heading` | the heading landing, and `p` from mid-file |
| `scroll::the_file_step_stops_at_both_ends` | that neither key moves the view backwards |
| `follow::a_file_step_disengages_follow_at_both_ends_of_the_diff` | intent, not movement, disengaging follow |
| `list::a_digit_jumps_to_the_file_on_that_row_of_the_window` | the window, not the changed set |
| `list::a_digit_past_the_drawn_window_is_a_no_op` | both bounds, in the two shapes that need them |

Three existing gates were extended rather than duplicated: `only_the_actions_that_move_the_viewport_disengage_follow`, `only_the_action_that_reads_the_height_is_given_one`, and `nothing_a_reader_did_not_ask_for_becomes_an_action`.

## The `/simplify` pass, and what it found

Four review angles ran in parallel. Three returned; the fourth (altitude) was stopped after fifty minutes still cycling mutations without producing a report, so that angle was judged by hand. Saying so because a review nobody reports is one nobody can tell you skipped.

**Applied:**

- **`App::jump_to`.** `anchored = false; position = {file, row: 0}` was written out at five sites and this change was adding a sixth. `App::scroll` already carries the argument for the other half of the pair, that a rule spelled out three times is one an arm eventually forgets, and the jump side never got it. The two reviewers split: one said extract, the other said the helper would serve two of five sites because `jump_to_newest` needs a flag and `Top`/`Bottom` build their position differently. Checked directly, and that is wrong on both counts. It serves five of five.
- The `File` arm becomes a let-chain, already the house idiom.
- `row_of` carried two fallbacks where one is reachable.
- `'0'` and `'7'` were asserted inert twice, once derived from `LIST_ROWS` and once hardcoded. The hardcoded pair is gone: raising the cap would otherwise redden a test named for idle cost, and the file's own convention is that keys with a dedicated test are not restated there.
- The short-pane fixture went from forty files to eight.

**Skipped, with reasons:** merging the new follow test into `jumping_to_the_last_file_disengages_rather_than_re_engaging` (they assert different rules, and it would put file steps under a name about `g`/`G`); swapping `materialise` for `frame.advance()` in two `list.rs` fixtures (the reviewer that raised it recommended against, since `materialise` is the settled convention at every fixture in that file).

## The coverage hole the pass found

A mutation added while re-checking the refactor deleted the `file < frame.files().len()` bound and **survived**. Both halves of `a_digit_past_the_drawn_window_is_a_no_op` are caught by the row bound first, so neither isolates it.

Reading then says the bound is dead, and that is the wrong conclusion. Given the row bound, no gesture against a freshly drawn list can name a file past the end, because the window is never taller than what is left below it. What the bound guards is the screen being **out of date**: `list_rows` is what the last frame drew, and the changed set can shrink before the next one. Six rows drawn, two files left, and `6` still passes the row bound.

`a_digit_after_the_diff_shrank_names_no_file` is the gate. It asserts on `App::position` rather than on a view, because a view resolves a degenerate position by clamping it and would hide the defect behind the clamp. `SPEC.md` §11.1 and the arm's comment now state what the bound is for instead of the reason reading suggested.

## Final gates

- `cargo test --workspace`: **573 passed, 0 failed**.
- `cargo test --release --test budgets` plus the release core gates: **22 passed, 0 failed**.
- `cargo clippy --all-targets` and `cargo fmt --all -- --check`: clean.
- **Twelve mutations, all killed**, including two written against the refactor itself (a jump that stops clearing `anchored`, and a jump that lands one row below the heading).


breferrari:
## The approved plan, carried over from #122

Commented on the issue at plan time (before any code), and repeated here so the fidelity check has both sides on one page.

---# #122. File-granular navigation: `n`/`p`, and digits `1`–`6`

Branch `file-navigation` in worktree `C:/Dev/vigia.b`, off `origin/main` (fb48275).

## Context

`vigia`'s keymap has row granularity (`j`/`k`), page granularity (`Space`/`PgUp`/`PgDn`/`d`/`u`)
and the two ends (`g`/`G`). It has no **file** granularity, and the file is the unit the pinned
list draws, the header counts, and the reader thinks in. Today, moving to the next file means
holding `j` through the remainder of the current one or reaching for the mouse. A mouse gesture
already exists for it (`Action::ListRow`, a click on a listed file), so the affordance is present
and only the keyboard route is missing.

Phase 8's filter is *"does it change what the pane looks or feels like in the first minute of
use?"*. This is input, not layout: no region moves, no chrome changes, no budget surface is
touched. Issue [#122](https://github.com/breferrari/vigia/issues/122).

## What the record gave me

Read before planning, and each of these constrains the design below:

- **`SPEC.md` §11.2 B4, ruled 2026-08-01 (#66): the list is a region and it is not navigable.**
  No selection, no focus, no second mode, and `▸` is a marker rather than a cursor. So `n`/`p`
  and the digits move the **diff**, exactly as a list click does; nothing is remembered and no row
  becomes special.
- **`SPEC.md` §11.1: "no key changes meaning depending on where anything is."** This settles the
  one genuinely open question below (what `p` does from mid-file) without needing a ruling.
- **`SPEC.md` §11.1 + #121, merged one commit ago (fb48275): the fifth hint does not fit.** At
  eighty columns the footer is exactly full: 40 columns of hints, 2, 19 of readouts, 2, 13 of
  state, so a fifth hint is paid for out of the frame-time and memory readouts.
  `legibility.rs::a_wider_hint_bar_cannot_quietly_push_the_readouts_out` pins both boundaries.
  #122's body already floats `n p files` only as a "bonus rung", and #121's ruling is one commit
  old. **The documented choice is taken: no hint bar change.**
- **`SPEC.md` §11.1: `Space`, `PgDn` and `PgUp` have never been hinted either.** An unhinted
  binding is an established family here, so adding three unhinted ones costs nothing new.
- **`SPEC.md` §11.1: `g`/`G` teach that case is load bearing.** `N`/`P` stay unbound for the
  reason `D`, `U` and `F` are.
- **`Action::is_manual_scroll` is an exhaustive match on purpose**, so a new action must rule on
  follow mode rather than inherit a default. Same for `Action::needs_height`.
- **Vault:** `search` over the vigil MCP returned `#66`'s two notes (the pinned region, the
  row-exact bar, B4 ruled) and nothing else touching the keymap. `recall` returned no memory
  about input, key choice or follow-disengagement. **There is no recorded decision this plan
  contradicts, and none it can lean on beyond `SPEC.md` itself**, and that is a finding, not a blank.

## Premises

| Must be true | Falsified by | Answer, and where it came from |
|---|---|---|
| `n`, `p` and `1`–`6` are unbound today | any of them already mapping | **Verified.** `input.rs::key_action` binds `q Esc j k J K Space PgDn PgUp d u g G Home End f` and `Ctrl-C`/`Ctrl-D`; every other code falls to `_ => None`. No digit is bound. |
| A digit can name a list row that is not drawn | the list region always being `LIST_ROWS` tall | **Verified by arithmetic on `Body::split`.** `affordable = body − (MIN_BODY + 1)`, `list = files.min(6).min(affordable)`. At a 9-row pane with a 1-line footer, `body = 7`, `affordable = 4`, so the list is **4 rows** with 40 files changed. Digits `5` and `6` then name nothing on screen. **A guard against `list_rows` is load bearing, not defensive.** |
| `App` knows how many list rows were drawn | the field not existing or being stale | **Verified.** `App::list_rows` is set from `body.list` in `App::view` on every frame, and `App::browse` already clamps against it for exactly this reason. |
| An out-of-range jump may disengage follow | it being a silent state loss | **Answered from precedent, not assumed.** `Action::Top` at the top and `Action::Bottom` at the last file already disengage follow while moving nothing: on this map follow yields to *intent*, not to movement. `follow ▶` leaving the footer is visible, unlike the map detaching, which is why `browse` has the opposite rule. #122 asks for a digit past the count to behave "the same as clicking empty list space", and that is what this preserves. |
| `usize::checked_add_signed` is available | MSRV below 1.66 | **Verified.** `rust-version = "1.85"`. |
| Only three sites match on `Action` | a fourth match failing to compile late | **Verified.** `grep -rn "Action::"` over `crates/*/src`: `app.rs::apply`, `input.rs::is_manual_scroll`, `input.rs::needs_height`. Nothing in `lib.rs`, `render.rs` or `view.rs` matches on it. |
| No snapshot changes | a `.snap` going stale | **Inferred, and gated below.** Nothing drawn changes: no hint, no region, no chrome. `cargo insta` acceptance is not expected and a stale snapshot would fail the suite loudly. |

## The design

### 1. `n` / `p`, a new action

`Action::File(isize)` in `crates/vigia/src/input.rs`, beside `Page`/`HalfPage`, which are the
existing "move by a unit larger than a row" variants and are also signed steps rather than paired
directional variants.

```rust
/// Move the viewport by this many changed files, negative for back.
File(isize),
```

- `key_action`: `KeyCode::Char('n') => Some(Action::File(1))`,
  `KeyCode::Char('p') => Some(Action::File(-1))`.
- `is_manual_scroll` → **`true`** (arm added to the `Scroll | Page | HalfPage | Top | Bottom` line):
  it moves the diff, so it disengages follow exactly as `G` does.
- `needs_height` → **`false`** (arm added to the `Scroll | Top | Bottom | ScrollList` line):
  it steps a file index, not a screenful.

`App::apply` arm:

```rust
Action::File(step) => {
    let files = frame.files().len();
    if let Some(file) = self.position.file.checked_add_signed(step).filter(|f| *f < files) {
        self.anchored = false;
        self.position = Position { file, row: 0 };
    }
}
```

**One rule: step the file index, land on the heading, do nothing when there is no such file.**
Three consequences, each deliberate and each stated in the doc comment:

- Row 0 is the heading, the same resolution a list click and `jump_to_newest` already use, and it
  costs no diff, and `I4` is untouched because nothing here asks how tall anything is.
- `p` from mid-file goes to the **previous** file, never to the top of the current one. §11.1's
  *"no key changes meaning depending on where anything is"* is the reason: the vim-style
  "top of this section first" behaviour would make one key mean two things, and `g` already
  reaches the top of the diff.
- `n` at the last file and `p` at the first are **no-ops for the position**, so neither key ever
  moves the view in the direction opposite to itself. Follow still disengages, per the precedent
  above.

`anchored = false` because this is a jump, not a scroll, the same reason `Top`, `Bottom` and
`ListRow` set it, and `App::anchored`'s own doc says why (a jump claims what belongs at the *top*,
so the viewport must not back up to fill a short tail).

### 2. `1`-`6`, reuse `Action::ListRow`, do not add a variant

`ListRow(u16)` is already documented as *"put the diff at the file this many rows down the pinned
list: an offset into the window, not a file index"*, which is exactly what #122 specifies for the
digits (*"what you can see is what you can name"*). A second variant would be a second spelling of
one intention, and both would need the same guard.

- `key_action`: `KeyCode::Char(digit @ '1'..='6') => Some(Action::ListRow(<digit − 1>))`.
- `'0'` and `'7'`–`'9'` stay **unbound**, returning `None`. That is a real distinction rather than
  an accident: an unbound key produces no action at all and so cannot disengage follow, where a
  bound one naming a row that does not exist is the "empty list space" case #122 calls for.
- The `'1'..='6'` range restates `render::LIST_ROWS`. `input.rs` is a pure function of a key code
  with no layout knowledge, and importing a render constant into it would give the crate an
  input→render edge it does not have. The restatement is made safe by a gate (below) that presses
  every digit `1..=LIST_ROWS` and asserts `LIST_ROWS + 1` is unbound, so raising the cap fails
  loudly instead of silently leaving a row unreachable.

`App::apply`'s `ListRow` arm gains the window guard:

```rust
Action::ListRow(offset) => {
    let file = self.list_top.saturating_add(usize::from(offset));
    if usize::from(offset) < self.list_rows && file < frame.files().len() {
        self.anchored = false;
        self.position = Position { file, row: 0 };
    }
}
```

**This changes nothing for a click**, because `input::Regions::over_list` already bounds a mouse row to
the region, and it is what makes the digits address the drawn window rather than the file list.
The doc comment records both halves.

### 3. `SPEC.md` §11.1

Three rows into the keys table, after `G`/`End` and before `J`/`K`:

| `n` | next changed file |
| `p` | previous changed file |
| `1`–`6` | put the diff at the Nth listed file |

Plus one paragraph under the half-page one, carrying: the file is the unit the list draws; the
jump lands at the heading for the reason a list click does; `p` never means "top of this file"
and §11.1's own no-key-changes-meaning rule is why; `n`/`p` at the ends move nothing; the digits
address the visible window and cap at `LIST_ROWS`; none of the three is hinted, on #121's ruling
that the footer is exactly full at eighty columns; `N`/`P` are unbound for the reason `D`/`U`/`F`
are; and B4 is untouched because all three move the diff and remember nothing.

### 4. `ROADMAP.md`

Flip the #122 row from `⬜` to `✅`.

## Tests

Every one of these fails against `origin/main`.

**`crates/vigia/tests/input.rs`** (the key map as a table):

- `n_and_p_are_the_file_step`: `n` → `File(1)`, `p` → `File(-1)`; `N` and `P` are inert, for the
  reason `D`/`U`/`F` are.
- `every_row_the_list_can_draw_has_a_digit`: for `row in 0..LIST_ROWS`, the digit `row + 1` maps
  to `ListRow(row)`; `'0'` and the digit after `LIST_ROWS` are `None`. This is the gate that makes
  the restated `'1'..='6'` range safe.
- `only_the_actions_that_move_the_viewport_disengage_follow`: extend the existing true-list with
  `Action::File(1)` and `Action::File(-1)`.
- `nothing_a_reader_did_not_ask_for_becomes_an_action`: extend the existing inert list with
  `press(KeyCode::Char('0'))` and `press(KeyCode::Char('7'))`.

**`crates/vigia/tests/scroll.rs`** (position against a real 40-file frame):

- `n_and_p_step_one_file_and_land_on_its_heading`: from `{file: 0, row: 0}`, `n` three times
  lands `{1,0} {2,0} {3,0}`; `p` twice lands `{2,0} {1,0}`. Then seed a mid-file position with
  `Scroll(2)` and assert `p` lands on `{file − 1, row: 0}` rather than on the current file's
  heading, which is the §11.1 ruling made falsifiable.
- `the_file_step_stops_at_both_ends`: `p` at file 0 does not move; `Bottom` then `n` does not
  move; and neither ever moves the viewport backwards.
- `only_the_action_that_reads_the_height_is_given_one`: add `Action::File(1)` and
  `Action::File(-1)` to the enumerated list, so the `needs_height` claim is checked against
  `App::apply` for the new variant too.

**`crates/vigia/tests/follow.rs`**:

- `a_file_step_disengages_follow`: with follow engaged, `n` disengages it, and so does `p` at
  file 0 where nothing moved (the intent-not-movement rule, asserted rather than left implied).

**`crates/vigia/tests/list.rs`** (the region, and the digit's window semantics):

- `a_digit_jumps_to_the_file_on_that_row_of_the_window`: 40 files, wide pane. Assert digit `3`
  lands on file 2; then `ScrollList` the window down and assert digit `3` lands on
  `list_top + 2`, which is what "addresses the visible window, not absolute indices" means and
  what a naive absolute-index implementation would fail.
- `a_digit_past_the_drawn_window_is_a_no_op`: the two shapes that produce one, because they are
  different defects: (a) **fewer files than rows**, 3 files, wide pane, digit `5` moves nothing;
  (b) **fewer rows than digits**, 40 files on a 9-row pane where `Body::split` gives a 4-row
  list, digit `6` moves nothing. Shape (b) is the one the new `list_rows` guard exists for and is
  the one that passes without it.

## Out of scope, stated

- **No hint bar rung.** #121's ruling, one commit old, and #147 owns the question underneath it.
- **No change to `soak.rs::scripted`.** Adding a `File` step would change what the soak measures,
  and the soak carries budget gates; #122's own Gates section does not ask for it.
- **No `Tab` cycling, no `Ctrl` binding, no selection or focus.** #122 refuses all three and B4
  is why.
- **No mouse change.** The click path already exists and its behaviour is unchanged.
- **No fix for "an out-of-range jump disengages follow".** It is pre-existing on the click path,
  #122 explicitly asks the digit to match it, and the precedent (`g` at the top) says intent is
  what disengages. Flagged in the report rather than changed under cover of this issue.

## Verification

Code diff, so the full gate runs, not the docs carve-out:

```sh
cargo test --workspace                 # every test, count named in the report
cargo test --test input --test scroll --test list --test follow   # the new gates
cargo clippy --workspace --all-targets -- -D warnings
cargo fmt --check
cargo test --release --test budgets    # frame path untouched, but the diff is code
```

Then the Phase 8 discipline: `/simplify` (the diff is small and feel-class, not engine-class),
a screenshot of the pane in the PR, and the step 6 plan-fidelity diff.

Issue #234, opened 2026-08-18
The sparkline's bucket ceiling was the band's period, and that period is gone


breferrari:
[#161](https://github.com/breferrari/vigia/issues/161) closed its sparkline half with a ruling rather than a rung: the sparkline was already at its ceiling, so there was nothing above `HISTORY_BUCKETS` to grow into. That ruling rested on a bound that no longer exists.

## The argument, and the part of it that died

The bound was the churn band. Both elements read one store over one window and a reader has both on screen, so a burst that splits in a file's sparkline and merges in the worktree band at the same instant is incoherent. The band drew a fixed `GRAPH_COLUMNS` of `GRAPH_PERIOD`, eight seconds, so a drawn sparkline bucket had to stay coarser than that: `HISTORY_BUCKETS` under fifteen, and twelve is the largest divisor of 120 that clears it.

[#232](https://github.com/breferrari/vigia/issues/232) redrew the band the way `btop` draws a graph. It plots one value per sub-column now, so its period is a property of the pane rather than a constant, and at any wide pane it is the store's own one second. `GRAPH_COLUMNS` and `GRAPH_PERIOD` are retired.

**The band is therefore finer than a drawn bucket at every width, and can never be coarser.** The constraint that made twelve a ceiling is gone. It was not replaced by a different bound; it simply stopped existing, so the ruling is reopened rather than defended.

## What is now open

Whether the sparkline's bucket count should become a width rung after all, which is what #161 asked for in the first place and what its heat-strip half now does.

- The window is 120 samples over 120 seconds, so the divisors above twelve are 15, 20, 24, 30, 40, 60 and 120: 8, 6, 5, 4, 3, 2 and 1 seconds a bucket.
- `FileEntry.spark` is `[u32; HISTORY_BUCKETS]`, so raising the source is a public type change, which is what raising it 8 to 12 already was.
- The layout side is already built: `SPARK_RUNGS` is a halving ladder and `Columns::plan`'s share clamp is what decides when a rung above the settled ladder is affordable. A new top rung is one entry, exactly as the heat strip's 48 will be.

## What argues against it

The measurement #223 took still stands as a measurement: distinct drawn heights peak near four to five seconds and collapse by fifteen. Twelve buckets is ten seconds, which is on the coarse side of that, so finer would likely read better and this is not merely a size question.

[...]

Issue #162, opened 2026-08-15
Side-by-side regions at real width: the list becomes a left rail


breferrari:
Rung 6 of [#125](https://github.com/breferrari/vigia/issues/125), the top rung.

## The gulf

On a ~210-column pane, paths end near column 40 and every glance element is pinned to the right edge, 150 cells away. The columnar slots keep rows comparable down the list ([#77](https://github.com/breferrari/vigia/issues/77)), but the **per-row** association, which strip belongs to which file, is destroyed by the void between them. The eye has to travel the whole pane to join a file to its own numbers.

Stretching one column does not fix that. At real width the answer is placing regions side by side.

## What it is

Past the width where one reading column stops being useful (~160?), the list stops being a strip above the diff and becomes a **left rail**: paths with their glance elements adjacent, the diff taking the remaining width as its own region.

It kills the gulf structurally rather than cosmetically, since the glance cluster sits beside its path rather than across a void, and it is the layout under which the multi-worktree deferral in §4 would eventually draw one rail per worktree.

## What transfers unchanged

The two-region model, the wheel-under-pointer rule, and the follow and caret mechanics. The regions move; their contracts do not.

## What it dissolves

[#124](https://github.com/breferrari/vigia/issues/124) ruled that the rule between the regions stays bare, and `Body::split`'s `rule: list > 0` makes the rule and the list coextensive. At this rung there is no horizontal rule at all, because the list is beside the diff rather than above it. #125 said this in its own body: that ruling is **dissolved** here rather than reopened.

## Gates

- Monotone across the transition. Widening a pane into the rail must not take an element away, and this is the hardest case the layout table has, because the two shapes allocate differently and the boundary is where they meet.
- `MIN_PATH_WIDTH` and I6's forty-column floor untouched. This rung exists above them and must not reach down.
- Both regions keep their scrollbars, their caret rules and their planning widths, including the column a bar costs whether or not one is drawn.
- I9 at the width where both regions draw a full complement at once.

## Blocked by

Nothing formally. It re-pins the layout table as rung 5 does, so the two should not run in parallel.

breferrari:
Split in two on 2026-08-21, before a plan was written rather than after an audit refused to converge, which is the same call [#125](https://github.com/breferrari/vigia/issues/125) made and the reason this issue exists at all.

[...]

Issue #125, opened 2026-08-05
The pane rewards space: density scaling, the churn graph, and side-by-side regions at width


breferrari:
## The pane rewards space: density scaling, the churn graph, and the cockpit at width

The layout ladder only ever *loses* going narrow; going wide it gains nothing, and a live run on a ~210-column pane shows what that costs, worse than any snapshot had:

- **The gulf.** Paths end near column 40; every glance element is pinned to the right edge, 150 cells away. The columnar slots keep rows comparable down the list, but the *per-row* association — which strip belongs to which file — is destroyed by the void between them. The eye has to travel the whole pane to join a file to its own numbers.
- **Postage stamps.** Twelve heat buckets and eight sparkline buckets are fixed sizes; on a wide pane they are specks. The elements that carry §5's whole differentiator occupy about 10% of the row that exists to display them.
- What already works at width: the full-bleed washes read perfectly (a clipped all-additions line banded edge to edge with its `›` mark is exactly right), and per-file counters stay legible.

A monitor worth the name scales the other way: more space means more drawn state — and at real width the answer is not stretching one column, it is placing regions **side by side**. Every rung below is **ink on frames that already happen** — zero wakes, zero writes, zero clocks, all of it drawing state the tool already retains.

Six rungs upward, in the order they pay:

1. **The counters take the mockup's colours.** `+42` green, `−7` red — the picture has drawn them that way from the start and §5.1 records the shell's grey as a departure that "still stands." It stops standing: this is compliance, not change. (Smallest item, listed first because it is the one with a spec debt attached.)
2. **The worktree churn graph — the hero element.** The hero element of any monitor is its big always-on graph; this tool's is worktree-wide churn over the I10 window, and the store **already holds it**: summing the per-path buckets is arithmetic on retained state. Drawn as a braille or block area band when the pane is tall enough to give it a row or two without costing the diff its `MIN_BODY` — the one glance answer nothing currently draws: *how hot is this worktree right now, and was it hotter a minute ago?*
3. **Braille resolution as a glyph rung.** The sparkline's eight buckets in four cells, or sixteen in eight, on terminals whose fonts carry braille — a rung above the block ramp exactly the way truecolour sits above 256. Ties into §10's open glyph-ladder question rather than pre-empting it.
4. **The list deepens on a tall pane.** Six rows is the cap a 24-row pane affords; a 50-row pane keeping six draws void where the map could be. The cap becomes a function of height with six as its floor, the same shape every width ladder already has.
[...]

Issue #204, opened 2026-08-17
The masthead starts drawn, and the reader who asked for the toggle wants it hidden


breferrari:
**The masthead is on by default, and the reader it was built for wants it off.** `m` shipped with the band in #158 and `SPEC.md` §11.1 rules the element on until a keypress takes it away. Reported from use the day after it landed: the band should start hidden, and `m` should be how a reader brings it up when they want it, which is the same gesture in the other direction.

**What the current ruling rests on is discoverability, and that is the whole of it.** The spec's words are *"It is on by default, because an element nobody discovers may as well not exist"*. B6 gives the tool no config file and no flags, nothing persists between runs, and `m` is deliberately not in the hint bar, so off by default means the band reaches only a reader who has been told about it.

**That is paid in the README rather than in columns.** The keys table there does not name `m` at all today, which is a gap #158 left whichever way this goes. Adding the row is what makes the band discoverable without spending readout columns at narrow widths, and the hint bar stays where #121 and #147 left it.

**What changes.** The shipped shell starts with the band hidden, `m` shows it and hides it again, and the spec paragraph that rules the opposite is amended with this issue as its citation. The toggle's app side has no gate today, which matters more once the default hides the element: a broken `m` would put the band out of reach entirely.

PR #249, opened 2026-08-20
The list deepens on a tall pane (#160)

### LINKED ISSUE #160: The list keeps six rows on a fifty-row pane, so the map is void where it could be
Rung 4 of [#125](https://github.com/breferrari/vigia/issues/125).

## What it is

`LIST_ROWS` is the cap a 24-row pane affords. A 50-row pane keeping the same cap draws void where the map could be. The cap becomes a function of height with today's number as its floor, the same shape every width ladder here already has.

## What it needs

- A height ladder for the list cap, with today's cap as the floor and `Body::split`'s existing clamp order unchanged: one row per changed file, then the cap, then whatever leaves the diff above `MIN_BODY`.
- Nothing else moves. The list already scrolls with `J` and `K`, and the digits already address the rows it draws.

## The digits are the question this rung has to answer out loud

`1` to `6` put the diff at the Nth row the list is drawing ([#122](https://github.com/breferrari/vigia/issues/122)). A cap that grows past six either leaves rows nobody can address by digit, or grows the keymap.

[#149](https://github.com/breferrari/vigia/issues/149) ruled that the list does **not** draw the rank the digits address, on the ground that a rank is a fact about the keymap where every other thing in that region is a fact about the worktree. That ruling is what makes unaddressable rows unremarkable rather than a hole, because nothing on screen ever promised the reader a number per row. It should be said rather than assumed, since it is the whole reason this rung does not drag the keymap behind it.

## Gates

- `MIN_BODY` inviolable, gated at the heights around the boundary.
- Monotone in height: a taller pane never draws fewer list rows.
- The floor holds: at 24 rows the list draws exactly what it draws today.
- `J`/`K` scrolling and the digit jumps both still land where they claim at the deeper cap.

## Blocked by

Nothing.

breferrari:
Closes #160. Rung 4 of #125.

**What a reader can now do that they could not before:** on a pane taller than 27 rows the pinned list draws more of the map. A 50-row pane draws twelve rows where it drew six, and the diff keeps three quarters of the pane at every height.

## The change

`LIST_ROWS` was a flat six, the cap a 24-row pane affords. The cap is a share of the pane now, floored at the number that shipped:

```rust
const LIST_SHARE: usize = 4;          // one row of list per four rows of pane

const fn deep_of(height: u16) -> usize { height as usize / LIST_SHARE }

const fn list_cap(height: u16) -> usize {
    let deep = deep_of(height);
    if deep > LIST_SETTLED { deep } else { LIST_SETTLED }
}
```

The branch rather than `.max` is the const context: `Ord::max` is a default trait method behind `const_cmp`, so no `const fn` in this file can call it. `Columns::plan` writes the same clamp as `.max` because it is a plain `fn`.

`Body::split` changes by one expression and the clamp order is untouched: one row per changed file, then the cap, then whatever leaves the diff `MIN_BODY`.

Written the way `Columns::plan` already writes its share one region out (`generous_of(width).max(SETTLED.width(glyphs))`), which is what makes "nothing below the new rungs can move" a mechanism rather than a claim.

### Why a quarter

Six is already a quarter of the pane it was chosen against, and that is not a fit. `SPEC.md` §11.1 derived it as *"on the 24-row pane this tool is built for it leaves fourteen rows of diff after the header, the rule and a one-line footer"*, and 24 less a header, the lead blank, six, the rule and a footer is fourteen exactly. The absolute (*"the largest block that still reads as a glance"*) was a proportion stated against a single height.

Checked outside the repo too: `bottom`, the tool `CLAUDE.md` names as this stack's precedent, sizes every layout component by `ratio` rather than by rows. Proportional regions are the monitor-class idiom.

| pane | 24 | 27 | 28 | 32 | 40 | 50 | 68 |
|---|---|---|---|---|---|---|---|
| list | 6 | 6 | 7 | 8 | 10 | 12 | 17 |
| diff | 14 | 17 | 17 | 19 | 25 | 32 | 45 |

The row a pane gains at a rung goes to the map; the other three go to the diff.

### The step of one is load bearing

The band is paid out of what the list leaves. `after = body - list - 2`, and one more row of pane adds one to `body` and at most one to `list`, so it never falls. A cap that gained two rows for one row of pane would take a band off a pane that had just grown, which is the "a bigger container holds less" failure the margin ladder is a written-out table to avoid.

### Read from the pane, never from the body

The ruling `margin_of` and `affords_caret` already make. The body's height depends on a footer whose own height depends on the changed-file count, so a cap read off it would deepen the list because a seventh file appeared.

## The two rulings

**The keymap does not grow.** The digits address the rows every pane drawing a list has, so `3` means the same thing at every height. Rows past six are reached with `J`/`K`, `n`/`p` and by clicking, which is bounded by the region rather than by the grammar and therefore reaches every drawn row.

One reason expired and is replaced rather than inherited: `7` to `9` were unbound because *"a row that can never exist should not spend a reader's follow mode"*, and a seventh row exists on any pane of 28 rows or more.

**#149 is re-ruled, and the rank is still not drawn.** §11.1 lists three things that reopen it, and the second is *"#125 drawing a list deeper than six on a tall pane"*. This rung is that, so the question was reopened rather than the answer repeated. Two reasons survived checking (a rank is a fact about the keymap where everything else in the region is a fact about the worktree; §5.3 prices attention rather than ink, which a deeper region strengthens), and the rung produced a new one that would settle it alone: with the digit grammar fixed at six, a rank on a deeper list would draw numbers on rows that are not keys. The condition is struck in place rather than deleted, the way B11's door was.

## Naming

`LIST_ROWS` is renamed `LIST_SETTLED`. It stopped being the rows a list takes the moment a rung existed above it, and `SETTLED`'s own docblock records what a stale layout name costs here. `LIST_FLOOR` was refused: `GRAPH_FLOOR` and `ROW_FLOOR` both mean *the smallest pane that draws the thing at all*, and this means the other end. Breaking on a `pub` export, so a **minor** bump.

`list_cap` is not exported. A test that imported the ladder would compare it against itself, which is why `HINT_RUNGS` is unexported too.

## Gates

- `the_list_deepens_on_a_tall_pane_and_keeps_its_settled_cap_below`: sweeps 120 heights. Monotone; steps by at most one row of list per row of pane; nothing at or below the reference pane moved; a 50-row pane draws deeper than six. Carries a witness that several rungs were reached, so a ladder that stopped laddering cannot pass by going quiet.
- `a_taller_pane_never_costs_the_band_its_rows`: the coupling the step bound protects.
- `the_list_region_gives_way_before_the_diff_falls_below_min_body`: widened from 40 heights to 120 rather than duplicated.
- `clicking_a_listed_file_sends_the_diff_to_it`: runs at 24 **and** 50 rows, which is what proves a row past the digit range is still reachable. It already clicks `list_rows - 1`.
- `every_row_the_list_can_draw_has_a_digit` renamed `the_digits_cover_the_settled_cap_and_stop_there`. Not every drawn row has a digit any more, and a test name that says otherwise is the quiet kind of wrong.

### Every gate proved red before it was trusted

Five mutants, reverted after each:

| mutant | result |
|---|---|
| flat cap, the pre-#160 behaviour | `the_list_deepens...` and the click gate red |
| a half instead of a quarter | `the_list_deepens...` red at `80x14`, and the existing cap gate red |
| two rows of list per rung | `the_list_deepens...` red at `80x16` on the step bound |
| cap jumps two rows **at** the band's arrival height | **green**, and that is a fact about the gate's reach |
| cap jumps two rows **one row above** it | `a_taller_pane_never_costs_the_band_its_rows` red at `80x25` |

The fourth is the interesting one. It left the band gate green because a jump placed exactly where the band arrives *delays* the band rather than removing it from a pane that had one. Delay is the ordinary shape of a clamp order and needs no gate; removal is what reads as a bug to the person watching. That reach is written into the gate's own docblock rather than left to be re-derived.

## What the review pass changed

Four review agents (reuse, simplification, efficiency, altitude), read-only, reasoning from the numbers above rather than their own runs. Four findings applied, one skipped.

- **The share was a pair and did not need to be.** It was `LIST_NUMER = 1` over `LIST_DENOM = 4`, defended in its own docblock as parity with `GLANCE_NUMER` / `GLANCE_DENOM` one region out. The parity is false: two fifths is irreducible and can only be written as a pair, where a quarter is `height / 4` exactly, so the numerator was an inert multiply and a name that did no work. Now `LIST_SHARE = 4`, written as the rate `SPEC.md` already states it in.
- **A docblock was nearly right, which is the kind that gets inherited.** It claimed "any denominator of two or more" holds the step bound the band rests on. A unit numerator holds it at any share at all, since `(h + 1) / n` exceeds `h / n` by at most one for every `n`. What four decides is which number the floor reproduces, not whether the step is safe.
- **A dangling `LIST_ROWS` in an OPEN `SPEC.md` §10 bullet**, left behind by the rename. It is more than a rename there: that bullet's bound was a constant and is the list's drawn height now, so its worst case grew with the pane it is measured on. Said so, since the bullet has to be re-measured when it is next picked up.
- **The read bound had a real coverage gap.** `SPEC.md` names `tests/reads.rs` as the authority for *"a visible list row costs one `Frame::diff` and no more"*, and every gate in that file measures 80x24, which is the one height where the list's share **is** the settled cap. So the claim was being held at exactly one rung of a ladder, and it was the rung that was already correct. That is the shape section 7 keeps finding, and this region has been bitten by it before. `layout_at` and `one_screen_at` take a height now, and `the_file_list_reads_only_the_rows_it_draws` holds the same claim a second time on a 50-row pane, where the region is twelve rows rather than six.

That last one got a sixth mutant, and it is the only one the old gate could not see: capping `take_list`'s loop at six rows however deep the region is drawn passes every assertion at 24 and fails at 50 with *"asked for 7 files, which is not one diff file plus 12 list rows"*.

**Skipped:** the review noted that `REFERENCE` in `tests/list.rs` names 24 while fourteen pre-existing sites in the same file spell it as a literal. They are not the same fact. Those fixtures want an ordinary terminal; `REFERENCE` is the ladder's anchor. Folding them would mean a future move of the anchor silently moved fourteen unrelated fixtures.

## Verification

- `cargo test --workspace`: **745 passed, 0 failed, 7 ignored**.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets` with `RUSTFLAGS=-D warnings` clean, which is the pair `ci.yml` runs.
- Budget gates green (13 in `tests/budgets.rs`), and nothing here touches the frame path: the change is one `min` argument in a pure layout function.
- `ROADMAP.md`'s row is flipped and the phase's paragraph is written.
- **No snapshot moved, and that was predicted rather than observed.** The tallest fixture anywhere in the suite is 24 rows, so every pane the suite draws is at the settled cap, and a rung added to the top of a monotone ladder cannot move a boundary beneath it.

## Scope

Docs and code both, so the full suite ran. Out of scope: #162 (side-by-side regions, which re-pins the layout table and must not run in parallel with this), #234, and any change to the digit grammar.

No ceiling on the ladder, deliberately. The region draws one row per changed file, so the cap binds exactly when a large changed set makes a bigger map worth more.


breferrari:
## What it looks like

Feel work, so here is the pane rather than a number. Rendered through the real `render` over a 30-file worktree, masthead off (the default since #204), 80 columns.

**80x24, the reference pane. Byte for byte what it drew before this branch:**

```
  vigia · main · 30 changed                                           watching

▸ M src/mod_0.rs                         ■■■■■■■■■■■■ ____________   +40   -40 ▲
  M src/mod_1.rs                         ■■■■■■■■■■■■ ____________   +40   -40 █
  M src/mod_10.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_11.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_12.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_13.rs                        ■■■■■■■■■■■■ ____________   +40   -40 ▼
  M src/mod_0.rs                         ■■■■■■■■■■■■ ____________   +40   -40 ▲
  @@ -1,40 +1,40 @@                                                            █
   1 - fn before_1() { let value = 7; }                                        │
   2 - fn before_2() { let value = 14; }                                       │
   3 - fn before_3() { let value = 21; }                                       │
   4 - fn before_4() { let value = 28; }                                       │
   5 - fn before_5() { let value = 35; }                                       │
   6 - fn before_6() { let value = 42; }                                       │
   7 - fn before_7() { let value = 49; }                                       │
   8 - fn before_8() { let value = 56; }                                       │
   9 - fn before_9() { let value = 63; }                                       │
  10 - fn before_10() { let value = 70; }                                      │
  11 - fn before_11() { let value = 77; }                                      ▼
  q quit · f follow · ? keys                                    follow ▶  1/30
```

**80x50. Twelve rows of map where there were six, and the diff still takes the rest:**

```
  vigia · main · 30 changed                                           watching

▸ M src/mod_0.rs                         ■■■■■■■■■■■■ ____________   +40   -40 ▲
  M src/mod_1.rs                         ■■■■■■■■■■■■ ____________   +40   -40 █
  M src/mod_10.rs                        ■■■■■■■■■■■■ ____________   +40   -40 █
  M src/mod_11.rs                        ■■■■■■■■■■■■ ____________   +40   -40 █
  M src/mod_12.rs                        ■■■■■■■■■■■■ ____________   +40   -40 █
  M src/mod_13.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_14.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_15.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_16.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_17.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_18.rs                        ■■■■■■■■■■■■ ____________   +40   -40 │
  M src/mod_19.rs                        ■■■■■■■■■■■■ ____________   +40   -40 ▼
  M src/mod_0.rs                         ■■■■■■■■■■■■ ____________   +40   -40 ▲
  @@ -1,40 +1,40 @@                                                            █
   1 - fn before_1() { let value = 7; }                                        │
   2 - fn before_2() { let value = 14; }                                       │
   3 - fn before_3() { let value = 21; }                                       │
   4 - fn before_4() { let value = 28; }                                       │
   5 - fn before_5() { let value = 35; }                                       │
   6 - fn before_6() { let value = 42; }                                       │
   7 - fn before_7() { let value = 49; }                                       │
   8 - fn before_8() { let value = 56; }                                       │
   9 - fn before_9() { let value = 63; }                                       │
  10 - fn before_10() { let value = 70; }                                      │
  11 - fn before_11() { let value = 77; }                                      │
  12 - fn before_12() { let value = 84; }                                      │
  13 - fn before_13() { let value = 91; }                                      │
  14 - fn before_14() { let value = 98; }                                      │
  15 - fn before_15() { let value = 105; }                                     │
  16 - fn before_16() { let value = 112; }                                     │
  17 - fn before_17() { let value = 119; }                                     │
  18 - fn before_18() { let value = 126; }                                     │
  19 - fn before_19() { let value = 133; }                                     │
  20 - fn before_20() { let value = 140; }                                     │
  21 - fn before_21() { let value = 147; }                                     │
  22 - fn before_22() { let value = 154; }                                     │
  23 - fn before_23() { let value = 161; }                                     │
  24 - fn before_24() { let value = 168; }                                     │
  25 - fn before_25() { let value = 175; }                                     │
  26 - fn before_26() { let value = 182; }                                     │
  27 - fn before_27() { let value = 189; }                                     │
  28 - fn before_28() { let value = 196; }                                     │
  29 - fn before_29() { let value = 203; }                                     │
  30 - fn before_30() { let value = 210; }                                     │
  31 - fn before_31() { let value = 217; }                                     ▼
  q quit · f follow · ? keys                                    follow ▶  1/30
```

The list's scrollbar is what shows the change is worth something: at 24 rows the thumb covers a fifth of the track and at 50 it covers two fifths, over the same thirty changed files. The caret, the sigils, the strips and the counts are the same drawer at both heights, because the region's row is one `Painter::file_row` and nothing about it is height-dependent.

Issue #322, opened 2026-08-25
The glance ramps become gradients and the band goes with them


breferrari:
The three glance ramps draw flat stops chosen by the 256-colour cube; at the truecolour rung the pane can afford btop-class gradients (#318 research: btop precomputes 101 steps per gradient as ready-to-emit colours, then colours per value for one-row elements).

## What

- Sparkline: per-bucket colour from a gradient over the value, staying in the cyan family (green means addition, SPEC §5.3, and the role rule was re-checked and holds).
- Heat strip: in-band gradients (green, yellow, red bands each ramp), keeping the three-band identity.
- Churn band: per-column gradient over height.
- Stops are theme keys; the mechanism is precomputed arrays, zero per-frame math beyond a lookup.

## Constraints

- At 256 the same stops quantise through the cube (btop's `round(c/51)` mechanism); at 16 colours today's stops stand; `NO_COLOR` unchanged.
- Budgets are tests: the frame path gains a lookup, and the gate proves it.

## Acceptance

- Tests pin the ramp per rung; screenshot in the PR against the current flat stops.

breferrari:
Plan, before code. The what is B18's; this is the diffable how, and one scope truth found by reading the code first.

## Premises

1. **The ramp mechanism is btop's**: linear RGB interpolation between the stops a theme already declares, computed once per frame per element (a few dozen lerps, no per-cell math). Verified in btop's source during #318; no easing, endpoints do the work.
2. **No new theme keys.** `spark`/`spark_warm`/`spark_hot` and the heat triples are already the `_start/_mid/_end` shape; the gradient interpolates between them, so every hue stays reader-movable and `docs/THEME.md` does not change.
3. **The depth ladder gates itself.** A resolved truecolour theme holds `Rgb` stops, which interpolate; a 256-resolved theme holds `Indexed`, which cannot, so those rungs keep today's three stops with no second mechanism. `ansi` and `NO_COLOR` unchanged.
4. **Scope truth: the heat strip does not gain a smooth ramp here.** Its projection carries three intensity bands per slice, not a height, so a gradient over it would be today's three stops wearing a new name. The elements that carry real height resolution are the sparkline (eight levels per bucket) and the churn band (eight levels per column), and that is where the ramp lands. Deepening the heat projection is core surgery the checkpoint did not price; it can be its own question if wanted.

## Design

- `theme.rs`: a pure `ramp` helper lerping three stops across N levels, returning the stops unchanged when any stop is not `Rgb`.
[...]

PR #329, opened 2026-08-26
The glance ramps interpolate in Oklab (#322)

breferrari:
The glance ramps interpolate (SPEC 11.2 B18, #322). Plan comment: https://github.com/breferrari/vigia/issues/322#issuecomment-5418642205

## What shipped

- `Theme::spark_ramp`: the three stop keys a reader already owns, interpolated into an eight-step ramp in **Oklab** rather than raw RGB, so a wide-hue ramp a custom theme asks for does not sag into grey-olive mid-way (lipgloss v2 blends in CIELAB for the same reason; Oklab is the cheaper, newer fit, inlined in ~40 lines rather than bought as a dependency SPEC.md does not name). Endpoints are byte-exact to the stops, and the gate proves it.
- Sparkline: per-bucket colour by value through the ramp, btop's one-row rule. The stop rides `Bucket` explicitly, computed from the same figure heights are scaled from, because reading it back off the glyph is lossy (the docblock that already knew that now gates a third consumer).
- Churn band: per-row colour against the vertical axis, btop's multi-row rule: quiet at the baseline, hot at the top, one lookup per row.
- The ladder gates itself: a theme resolved below truecolour holds indexed or named stops, which have no line to walk, so `spark_ramp` is `None` and everything draws exactly as it did. No flag, no second mechanism. `ansi` never ramps by construction.

## Scope truth, recorded in the plan before code

The heat strip does not gain a smooth ramp: its projection carries three intensity bands per slice, not a height, so a "gradient" there would be today's three stops renamed. The ramp lands where height resolution actually exists. Deepening the heat projection is core surgery the checkpoint did not price.

## Found while capturing

The first capture ran under the reader's own `~/.config/vigia/theme` and its cyan-to-red ramp went muddy in the middle, which is what moved the interpolation from RGB to Oklab before shipping: the customizability qualifier means wide ramps are reachable by configuration, so the mechanism has to survive them.

gradients over the delta washes, dark theme

Status bar in the capture reads 2.7ms a frame against the 16ms budget, live under a write burst.

## Tests

- `the_spark_ramp_interpolates_only_where_stops_are_rgb`: endpoints exact, 256-resolved and `ansi` refuse.
- `the_band_climbs_the_ramp_toward_the_top`: read off cells, mutation-checked by inverting the row index (fails as it should).
- `a_taller_sparkline_bucket_is_drawn_hotter` learned the ramp: rank comes from ramp position where one exists; the claim was always order, not identity.
- Full shell suite 31 binaries green, core 13 green, clippy clean, budget gates in the run.


breferrari:
Honesty note on the CI history: the first macOS failure was crates/vigia-core/tests/watch.rs::git_object_churn_never_produces_a_tick, which no line of this diff can reach, so it was an unrelated platform flake and the rerun is green. The endpoint-exactness commit stands on its own reasoning (byte-exact ramp ends whatever a libm rounds), not on having fixed that failure. Filing the flake observation on the workload-evidence issue rather than pretending it did not happen.

Issue #326, opened 2026-08-25
A path is a link: OSC 8 on the list and the headings


breferrari:
A monitor's paths should be one gesture from the editor. OSC 8 hyperlinks cost no column and degrade silently everywhere (text renders, link inert), and ratatui 0.30.1 added the cell hook that makes them safe under buffer diffing (`CellDiffOption::ForcedWidth`; `tui-link` and Codex CLI are prior art, #318 research section on OSC 8).

## What

- File paths in the list and in diff headings emit OSC 8 links to `file://` URLs; a config key can template the URL (delta ships `vscode://file/{path}:{line}` templating).

## Constraints

- Emit unconditionally: the matrix shows silent degradation everywhere, with Konsole shipping the feature off by default and losing nothing but the link.
- No column is spent; the path renders byte-identically without link support.

## Acceptance

- A cell-level test that the emitted sequence wraps the path and the diff stays width-honest; manual click check in foot and ghostty recorded in the PR.

breferrari:
Plan, before code.

## Premises, checked

1. **The cell hook exists in the shipped dependency**: `CellDiffOption::ForcedWidth` is in ratatui-core 0.1.2 in this workspace's registry cache, read directly.
2. **The proven mechanism is one cell, not per-cell wrapping**: tui-link concatenates open-sequence, URI, text and close-sequence into a single cell's symbol and forces the text's width (source read). Codex CLI ships the same shape. Adopted as-is.
3. **The worktree's absolute path is available where the session is built** (`worktree.workdir()`, lib.rs), and the display name derives from it already.
4. **Degradation is silent everywhere** (the #318 matrix): a terminal without OSC 8 renders the text and swallows the link; Konsole ships it off and loses only the link; tmux passes it since 3.4 and older tmux drops the OSC whole.

## Design

- `Chrome` gains `root` (absolute worktree path) and `links` (the config key). An empty root draws no links, which also keeps every hand-built test fixture linkless with no flag to remember.
- The config file gains `links`, **default on**, which makes it the first key whose default is not off: the issue's constraint is emit-unconditionally because degradation is silent, so the key exists to turn a nicety off, not to discover it. `Config`'s `Default` becomes manual and the docblock carries the ruling.
[...]

Issue #325, opened 2026-08-25
The pane takes its colours from the terminal, and follows a theme flip live


breferrari:
The defining 2025-2026 trend in terminal UI is taking colours from the terminal instead of shipping them at it (#318 research: delta's OSC 10/11 detection, yazi's OSC 11 + CSI ?996n + mode 2031 subscription, OpenCode's system theme generated from the terminal's own palette).

## What

- On request (a theme name like `system`, or a config default), query OSC 10/11 for fg/bg and OSC 4 for the 16-colour palette, classify light/dark by luminance, and generate the full theme from the terminal's own colours, diff washes tinted from the terminal's green and red.
- Subscribe to colour-scheme change notifications (mode 2031, DSR 997) so a monitor that sits open for days follows a daytime theme flip live.

## Constraints

- `VIGIA_THEME` with an explicit theme wins outright, today's behaviour.
- Queries die at a tmux boundary and on terminals that never answer: a timeout falls back to the current default theme, and the failure is silent by design.
- I1: the subscription is an event, not a poll.

## Acceptance

- Tests with a scripted pty answering and not answering the queries; light/dark flip exercised; documented.

breferrari:
Plan, before code, with one dependency-blocked half carved out loudly rather than absorbed.

## Premises, each checked against source

1. **crossterm 0.29 cannot deliver a colour-scheme notification.** Its unix parser (`event/sys/unix/parse.rs`, read in the registry cache) returns an error on any sequence it does not know and clears the buffer, so a mode 2031 / DSR 997 report arriving mid-session is silently destroyed before the app can see it. The **live flip is therefore blocked by the dependency**, not declined: it gets its own issue carrying this evidence, and the day crossterm surfaces unknown sequences (or vigia reads the tty itself) it becomes buildable.
2. **A startup query is not blocked**: before the event reader exists, the tty can be asked OSC 11 directly. `rustix` is already in-tree beneath crossterm and notify; one line in SPEC section 6 names it for this use, in its own commit, which is the house process for a dependency.
3. **The luminance rule is the field's**: OpenCode classifies the OSC 11 reply at 0.299/0.587/0.114 against 0.5; delta and yazi agree in mechanism. Reply forms `rgb:RR/GG/BB` through `rgb:RRRR/GGGG/BBBB` are all parsed, scaled by digit width.
4. **The OSC-4 synthetic palette is separable** and carved to its own issue: what changes the shipped default is detection, not palette synthesis.

## Design

[...]

Issue #324, opened 2026-08-25
The glyph ladder learns which terminals draw octants natively


breferrari:
SPEC said octants are unreachable because no font carries them. That was a fact about fonts quoted as a fact about rendering: ghostty 1.2+, foot 1.20+, kitty 0.40+ and VTE 0.78+ rasterise the octant and sextant ranges themselves, font-independently, exactly as every terminal rasterises box drawing (#318 research, terminal matrix, verified against each terminal's source or release notes; probe screenshots on this machine confirm foot and ghostty).

## What

- `Glyphs::detect` learns a terminal table (the same mechanism `Depth` already uses) and returns the octant rung on the four engines that draw it natively.
- A sextant rung where an element can use 2x3 density; Konsole is the one modern terminal that does not self-draw sextants, and the table says so.
- ratatui 0.30 ships `Marker::Octant` and `Marker::Sextant`, so the drawing side exists already.

## Constraints

- Under tmux, glyph width tables can lag (tmux 3.6+ `codepoint-widths`); the table stays conservative behind a multiplexer.
- `VIGIA_GLYPHS` still overrides everything, both directions.

## Acceptance

- Detection tests per terminal identity; the dense-rung gates extend to the new rung; screenshot on foot or ghostty in the PR.

breferrari:
Plan, before code. The what is B18's; the mechanism is the one the SPEC section-10 amendment already names: the table `Depth` and `Glyphs` always used, extended with the engines that draw octants themselves.

## Premises, each checked

1. **Four engines self-draw U+1CD00**: ghostty 1.2+, foot 1.20+, kitty 0.40+, VTE 0.78+ (the #318 matrix, verified per-terminal against source or release notes).
2. **ghostty advertises its version**: `TERM_PROGRAM=ghostty`, `TERM_PROGRAM_VERSION=1.3.1-arch2` (probed on this machine, Ghostty 1.3.1). Parse the leading `major.minor`, gate at 1.2.
3. **kitty**: `TERM=xterm-kitty`; `TERM_PROGRAM_VERSION` is parsed when present and gated at 0.40; when absent the answer stays braille, which fails safe whichever release introduced the variable (the web check could not pin it, so the design does not depend on it).
4. **VTE**: `VTE_VERSION` is numeric (`7802` is 0.78.2), ancient, and documented; gate at 7800.
[...]

Issue #323, opened 2026-08-25
The chrome earns 2026: segmented bars, a spliced sheet title, opt-in icons


breferrari:
Chrome polish from the #318 vision board (variant d), all font-guaranteed by measurement: the Powerline separator set is 16/16 and the icon planes are fully covered in the stock JetBrainsMono Nerd Font, and every serious tool ships a clean glyphless fallback (research section 1.3).

## What

- Header and status bar draw as segmented chips with Powerline caps (E0B0 family), colours as theme keys.
- The gestures sheet takes rounded corners and a border-spliced title (btop's `┐ title ┌` technique).
- File-type icons in the list, opt-in (off by default), theme-driven, with the plain rendering as the fallback (lazygit/yazi shape).

## Constraints

- Nothing costs a column that the current layout does not already spend (the #316 rule).
- Every glyph outside CP437 is stated against SPEC §10's Windows bullet, as that bullet requires.
- Icons off means byte-identical rendering to today.

## Acceptance

- Tests for the ladder (glyphs degrade per rung); screenshot in the PR.

breferrari:
Plan, before code. Three pieces, one decision taken against the issue's letter and stated first.

## The separator decision, taken now with its reason

The issue names Powerline caps (E0B0), measured 16/16 in the stock font. But a PUA glyph is a **font** bet, coverage is unprobeable from inside a terminal (#316's lesson, unchanged by B18), and chrome that drew tofu for a reader on an unpatched font would be the staged-mark mistake re-shipped as decoration. The chips ship with **block-element caps** (`▐` `▌`), which are in the range every font carries whole and every terminal self-draws, and which read as clean squared seams. The pointed caps become reachable the day a capability table can vouch for the font, which is not today. This is the plan's one departure from the issue text, taken before code, and the PR will carry it.

## Design

1. **The gestures sheet**: rounded corners `╭╮╰╯` with the title spliced into the top border as `┐ gestures ┌` (btop's technique, read from its source in #318). Rounded rides the glyph ladder: the Block rung keeps square corners, because the legacy console's CP437 has no arcs and that bullet's discipline in SPEC section 10 is per-glyph honesty. Geometry unchanged, so the sheet's row/width gates stand as they are.
[...]

PR #338, opened 2026-08-26
The README's ladders and the picture catch up (#318)

breferrari:
Four statements in `README.md` stopped being true when the B18 slate landed, and the picture the README leads with stopped agreeing with the binary. This is the correctness half; the missing-feature prose follows separately.

## What was false

The precedence table never moved when the ladders under it did, and two of its rows contradicted prose in the same document:

| Row | Said | Actually |
|---|---|---|
| Palette | `VIGIA_THEME` → theme file → `ansi` | the terminal's own background sits between the file and `ansi` (#325) |
| Glyphs | ... → `TERM_PROGRAM` → `WT_SESSION` → `TERM` → braille | the octant rung was missing entirely (#324) |
| View | `~/.config/vigia/config` → everything off | `links` starts on (#326) |

The palette row contradicted the paragraph forty lines below it, which already said `vigia` asks the terminal at startup; the view row contradicted the config section, which already said `links` is the one key that starts on. A reader got both stories in one file.

## The picture, which SPEC section 5.1 makes a specification

`assets/preview.svg` is the README's hero image and the spec for the look, and the "About that picture" section makes a specific, checkable claim: *one* departure from the binary is left. Two more had appeared:

1. **The header's pills were missing.** #323 changed what the header draws at truecolour and section 5.3 requires the mockup to be corrected in the same pass. It was not. That is my omission from that pass, fixed here: the loud pill on the first fact, the quiet one on the rest, the separators left bare on the bar between them, painted before the text so the words sit on top.
2. **The sparklines were green.** The binary has drawn them in cyan since the ramp landed, on the ruling that green already means *added* two rows down, and the picture was never brought forward. The 24 bars now use the binary's own stops, computed from the same eight-step Oklab interpolation `Theme::spark_ramp` produces rather than eyeballed: `#39c5cf`, `#5cd7df`, `#7ae9f0`, `#8aecf2`, `#a8f2f7`.

**This is a visible change to the front-page image**, which is why it is called out here rather than buried: if the cyan is not wanted, the ruling to revisit is the one in section 11.1 that made the binary cyan, and this commit is the cheap thing to revert.

With both brought forward the section's claim is true again: the one departure left is the header reading the worktree's name rather than `vigia`. The enumeration in that section gained the pills and the gutter tones.

**Not changed:** the ASCII pane diagram. It is a layout diagram, the layout did not move, and it never could show colour. The wash and gutter *values* still differ from the binary's, which section 11.1 records as intentional: the picture specifies that a row is washed, not the exact triple.

## Checks

Docs and one asset; no code touched. `render` (130) and `legibility` (65) run green anyway, since those are the suites that read the drawn screen the picture claims to describe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,10 +311,10 @@ Three independent settings decide how the pane is **drawn**, and most confusion 
 
 | | First answer wins |
 |---|---|
-| 🎨 **Palette** | `VIGIA_THEME` (a name, or a path) → `~/.config/vigia/theme` → `ansi` |
+| 🎨 **Palette** | `VIGIA_THEME` (a name, or a path) → `~/.config/vigia/theme` → **your terminal's own background** → `ansi` |
 | 🔦 **Depth** | `VIGIA_COLOR` → `NO_COLOR` → `TERM=dumb` → `COLORTERM` → `TERM_PROGRAM` → `TERM` → 16 |
-| ✏️ **Glyphs** | `VIGIA_GLYPHS` → `TERM=dumb`/`linux` → `TERM_PROGRAM` → `WT_SESSION` → `TERM` → braille, or blocks on a bare Windows console |
-| 🪟 **View** | `~/.config/vigia/config` → everything off |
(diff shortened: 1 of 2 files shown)


breferrari:
Folded into #341, which fixes the rendering defects reported from a real pane and carries these README corrections with them. The picture's pills are removed there rather than added: they were reported as reading badly, so #341 takes them out of the binary and the mockup together.

Issue #318, opened 2026-08-25
What a monitor-class TUI can draw that vigia is not drawing


breferrari:
Asked for from the pane while [#316](https://github.com/breferrari/vigia/issues/316) was being built, and it is the more interesting half of that pass.

`btop` installs on all three tier-1 platforms and looks far better than this. `#316` refused a combining-mark overlay on font coverage, and the measurement that settled it also says where the ceiling actually is, which is much higher than this tool currently reaches.

## The measurement

Sampled with `fc-list ':charset=<hex>'` against a stock `JetBrainsMono Nerd Font`:

| range | coverage |
|---|---|
| Block Elements `U+2580..U+259F` | 32/32 |
| Box Drawing `U+2500..U+257F` | 32/32 |
| Braille `U+2800..U+28FF` | 16/16 |
| Combining marks for symbols `U+20D0..U+20FF` | 0/48 |

`btop` is rich **because** it stays inside the first three, which are CP437 heritage every monospace font treats as mandatory, and spends its ambition on truecolour instead. It never bets on a rare codepoint. That is a licence, not a limit: the safe vocabulary is much larger than what this pane uses today.

## What is worth pricing

- **Truecolour gradients per cell.** This is the whole of what makes other TUIs look modern, and it is just 24-bit foreground. `Depth::Truecolor` already exists and the heat strip and sparkline are the obvious candidates.
- **Braille for sub-cell density.** Eight dots per cell against the sparkline's current one glyph per sample. It is the standard trick for terminal graphs and the coverage is there.
- **Half and quarter blocks** for sub-cell vertical resolution, which is how `btop` draws a smooth curve in a character grid.
- **Underline styles**: curly, dotted, dashed, and the separate underline *colour* extension. `ratatui` has an `underline-color` feature already. Note §5.3 currently reserves underline for the pointer's channel, so this needs a ruling before it needs an implementation.
- **The glyph ladder already exists.** `Glyphs::detect()` degrades on `TERM`. Anything adopted here has to say what it becomes at the ASCII rung and at `Depth::Ansi16`, where backgrounds are dropped entirely.

## What is out of scope

Anything needing a codepoint outside the ranges above, for #316's reason: glyph coverage cannot be probed from inside a terminal, so it cannot be degraded around. And anything that costs a column, which is the specific mistake #316 undid.

[...]

PR #339, opened 2026-08-26
The README describes what the slate added (#318)

breferrari:
Stacked on #338, which fixes what the README got *wrong*. This adds what it never said.

The diff presentation is the most visible thing the B18 slate shipped and the README described none of it: a reader was told the diff is syntax highlighted, which has been true for months, and nothing about the surface that actually changed under them.

## What was added

- **A section on the line itself**, beside the highlighting one: the calm wash, the hotter patch on the words that actually changed, and why the pairing is bounded rather than greedy (a pair too far apart is left as two whole lines, because past that bound there is no shared shape left to point at). The two-tone gutter gets its sentence. All three are backgrounds, so the paragraph ends where they end: below 24-bit colour the sigil column and the left bar carry the signal as they always did.
- **The header's pills**, one line where the pane is first described.
- **The sparkline's ramp**, one clause in the block that already explains what its height means: height and ink climb together, so the shape reads and the colour confirms.
- **The path is a link**, in the row table, where a reader looking at the row learns what the row can do.

## What was deliberately not added

**Icons stay in the config block only.** They are off by default, so a reader of "Reading the pane" would be shown something their pane does not have; the place to learn about an opt-in appearance key is the place you go to turn things on. If you would rather they were advertised where the row is described, that is a one-line move.

No promises, budgets or invariants were touched: this is description of what shipped, not new claims about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,8 @@ cargo install --git https://github.com/breferrari/vigia vigia
    status  │  q quit · f follow · ? keys   3.1ms frame   25MiB  follow ▶  1/3
 ```
 
+The header's facts sit in **pills**, one to a fact, so the worktree, the branch and the count read as three things rather than one run-on line.
+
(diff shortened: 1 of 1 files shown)


breferrari:
Folded into #341, minus the pills line: the pills are removed there.

PR #342, opened 2026-08-26
A file is never drawn without the label of the run it belongs to

breferrari:
## The report

The list drew a **staged** file on its last visible row while the only heading above it said `unstaged`, so the list filed the reader's change under the wrong run.

## The cause

`plan_with` in `crates/vigia/src/view.rs` dropped a run's label whenever the window had room for one more thing, and spent that row on the file instead. That was deliberate: furniture gives way before content does, the same rule the gutter and the caret follow.

That rule is wrong here, and the reader overruled it. The two outcomes are not comparable. A list one row shorter is merely smaller, and the rail already says there is more to come. A file under the wrong run's heading is the list asserting something false about where a change lives.

## The fix

The label is drawn before any of its run's files, without exception, and the file follows only if a row is left.

The degenerate case the old rule was guarding is still closed, in the other direction: a one-row grouped window now draws the label and no file. It says less than it could, but nothing untrue, and it never comes out empty (which was the original defect, where the label was emitted and then retracted).

## Gates

`no_file_is_ever_planned_without_its_own_runs_label` sweeps every `(top, rows)` the two-run fixture reaches and asserts each planned file sits under a label of its own origin. Watched red first, on the reported shape:

```
at top 0 with 1 rows, file 0 is Unstaged but the label above it says None
```

`a_one_row_list_draws_a_file_rather_than_a_label` encoded the overruled rule, so it is rewritten to the new one rather than deleted, and still pins the empty-list defect it was written for.

Verified in a real terminal at the tightest squeeze, where the boundary meets the bottom edge:

```
  ──  unstaged  7 ──────────────────────────────────────── ▲
  M f6.md                                       +1    -0 █
  M f7.md                                       +1    -0 │
  ──  staged  2 ────────────────────────────────────────── ▼
```

Before the fix that last row was `M f8.md` under the `unstaged` heading.

## The separate arrow report, still open

The other report (the continuation mark going out of line while scrolling) is **not fixed here, and not reproduced.** This PR strengthens the loop hunting it rather than guessing at a fix:

- it now drives real syntax spans, which was a genuine gap, since the mark is decided while walking spans and a fixture with none never reached the deciding code
- it asserts every mark in a frame shares one column, which states the reported symptom directly rather than a proxy for it

It passes across 40 body shapes and 35 scroll positions through the real two-buffer diff, and a 22-position sweep against the actual binary in a real terminal put every mark in column 74 without exception, emoji lines included. So the divergence is not in the buffer model. Left open.

Full workspace suite green.

### DIFF
--- a/crates/vigia/src/render.rs
+++ b/crates/vigia/src/render.rs
@@ -8571,6 +8571,23 @@ fn printable(text: &str, column: &mut usize, room: usize) -> Printed {
         .saturating_mul(CHARS_PER_COLUMN)
         .saturating_add(TAB_STOP) as u64;
     let mut examined = 0u64;
+    // Where the grapheme being measured began, and what it has cost so far.
+    //
+    // **A column advance belongs to the grapheme, not to the characters in
+    // it.** `ratatui` places `⚠️` as one two-column grapheme, while its two
+    // `char`s measure 1 and 0 apart, because the variation selector `U+FE0F`
+    // is what asks for the emoji presentation and carries no width of its own.
+    // Summing per character therefore undercounted the row by one column per
+    // emoji, and every column after it was wrong: the row overflowed the pane,
+    // the clip decision was taken against a column that did not exist, and the
+    // painter dropped characters to fit a row it had already mismeasured.
+    //
+    // Re-measuring from the cluster's start whenever a zero-width character
+    // arrives is what agrees with `ratatui` again, and it covers combining
+    // marks and ZWJ sequences by the same rule rather than by naming them.
+    // ASCII stays off the measuring path: it opens a cluster and pays nothing.
+    let mut cluster = 0usize;
+    let mut cluster_width = 0usize;
     for (i, c) in text.char_indices() {
         if *column >= room || examined >= walk {
             // Stopped with source left over, which is the caller's signal to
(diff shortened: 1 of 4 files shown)


breferrari:
## The arrow report is now root-caused and fixed

The repro Brenno gave (the real file, more files below it, a short narrow window) reproduced it immediately, and the cause is not where I was looking.

**It is `ratatui-core` 0.1.2, and it is triggered by `U+FE0F`.**

`U+FE0F` asks the terminal to draw the character before it as a two column emoji, and ratatui claims a second cell for the pair. Its buffer diff has a dedicated path for that second cell which emits it whenever its symbol changed, **without moving the cursor first**, on the assumption the terminal has not advanced past it. A terminal that drew the emoji two columns wide has advanced past it, so the write lands one column late and pushes the rest of the row right. The row then overflows the pane, the clip is decided against a column that is not there, and the continuation mark is drawn twice.

Confirmed on the wire rather than by reading. The bytes vigia emits around the emoji are:

```
`  ⚠️  ESC[39;49m  <space>  ESC[38;2;153;255;228;48;2;18;41;29m  <space>  **Gym
```

A bare style reset and a space directly after the emoji, with no cursor move between them, where the source has a single space.

Before, at 80 columns:

```
47 +   ? `⚠️  **Gym cashback reminder**: ${month} row in [[${relPath}]›█
48 +   : `⚠️  **Gym cashback reminder**:no ${month} rownyet in [[${re››│
```

After:

```
47 +   ? `⚠ **Gym cashback reminder**: ${month} row in [[${relPath}]]› │
48 +   : `⚠ **Gym cashback reminder**: no ${month} row yet in [[${rel› │
```

Spaces intact, one mark per row, marks aligned.

`ratatui-core` 0.1.2 is the current release, so the fix is to not reach the path: the presentation selector is dropped. That costs the text presentation of the glyph rather than the emoji one, which a monitor can afford where a shifted row cannot. Characters that are wide *without* a selector (CJK, most emoji) take ratatui's ordinary wide-character path, which is correct, and are untouched. That is also why my earlier CJK property sweep stayed green and never found this.

## Why the earlier loops all passed

They were sound and they were testing the wrong side of the seam. `TestBackend` records cells; it never replays the diff against a terminal that advances a cursor. Under the defect the buffer is **correct** and only the emitted byte stream is wrong, so every rendered-row assertion drew a correct row and proved nothing. The gate is therefore at the buffer, asserting the selector never reaches a cell, and the docblock records that reasoning. Watched red before the fix and green after.

Also fixed alongside: a column advance is now a grapheme's rather than the sum of its characters', which is what ratatui's own `cell_width` measures. Per character the pair measures 1 and 0 while together it measures 2.

Full workspace suite green.

Issue #380, opened 2026-09-02
Build B20: a drag over the diff selects rows, and `y` sends their real text


breferrari:
Build for the ruling in [#177](https://github.com/breferrari/vigia/issues/177). `SPEC.md` §11.2 B20 rules that the shell draws its own selection over the diff and that the copy is resolved against the row model rather than read out of the cells. Filed apart so the ruling does not wait on an implementation.

## What a reader gets

A left drag over the diff body washes every row it crosses. `y` then sends those rows' real text: the whole line where the pane clipped it, one whole line where wrap split it over several rows, the repository-relative path where the row is a file heading, the `@@` line where it is a hunk header. With no selection standing, `y` is unchanged and sends the caret file's path.

## The seams that already exist

- **The press, drag and release bracket.** `Held` from [#166](https://github.com/breferrari/vigia/issues/166) proved the lifecycle, and `hover_after` is the shape a pointer-state helper takes here.
- **The clipboard.** `clipboard::copy` is hand-written, declares no crate, and walks bytes rather than characters, so a multibyte line survives. `App::take_yank` to `Session::send` already carries a string of any length.
- **The real text.** `view::Row::Line` holds the whole line and `Row::Wrap` the whole tail, both independent of what the painter clipped.
- **The pointer state.** `Chrome` already carries `pressed`, `gripped` and `hovered` from the loop, and the diff row loop already knows each row's `y`.

## The change

- `input.rs`: a `Selection` of anchor and head screen rows with an ordered inclusive span, and `selection_after`, modelled on `hover_after` beside it. A `Down(Left)` inside the diff region and off its bar opens one, a `Drag(Left)` moves the head clamped to the region, `Up` leaves it standing, and a lost focus ends the gesture the way `Held::ends` does. No new `Action`: a selection is pointer state, not an intention.
- `lib.rs`: the selection on `Shell` beside `hovered`, cleared where an action is applied unless the action is `Yank`. One place, one rule, so the set of keys that clear is not a list anybody has to maintain.
- `app.rs`: the span onto `Pointing` and `Chrome`; `App::view` resolves it into the text right where `caret` is set after the collect, total over `Row`, with a wrapped line emitted once; `Action::Yank` prefers it and falls back to the caret path unchanged.
- `render.rs`: the wash in the diff row loop, one new theme token, and a ninth mouse gesture at both the wide and the tight spelling.
- `README.md`, `docs/THEME.md` and `ROADMAP.md`.

## Gates

[...]

PR #381, opened 2026-09-02
The shell draws its own selection, and the copy comes from the row model

### LINKED ISSUE #177: Text cannot be selected or copied
Requested from use: text on screen cannot be selected or copied, and it should be able to be.

## Why it does not work today, which is not an omission

The shell takes the mouse. `crossterm`'s `EnableMouseCapture` writes `\x1b[?1000h\x1b[?1002h\x1b[?1003h\x1b[?1015h\x1b[?1006h`, and while that is on the terminal emulator routes button presses and drags to **this program** instead of performing its own native selection. That is the same bundle `RULINGS.md`'s I1 section documents for a different consequence, and it is what buys the wheel, the draggable thumb, click-to-file and the step buttons. So selection was not overlooked: it was spent.

## There is a zero-code half, and it should ship first

Every major terminal keeps a modifier that bypasses mouse reporting and gives back native selection: **Shift** on xterm, GNOME Terminal, Konsole and Windows Terminal; **Option** on iTerm2; **Shift** on Kitty and WezTerm. A reader holding it can already select and copy from `vigia` today, with the terminal's own selection and the terminal's own clipboard, and nobody has been told.

That is a README line and possibly a hint, and it is worth landing before any of the below, because it is free, it works over SSH, it works in tmux, and it produces exactly the text the terminal has (which is the text on screen, elided marks and all).

## The in-app half, and what it costs

Doing it properly means the program owns the selection:

**The gesture.** Press, drag, release, with the selected span tracked between them. The press/hold/release lifecycle now exists: [#166](https://github.com/breferrari/vigia/issues/166) added `Held`, which arms on a press, carries the cell it began on, survives a twitch, and ends on release, on a key, on a pointer move, or on leaving the control. **That part is reusable and was built to be** (`Action::repeated` is the seam that makes the mechanism general rather than scrollbar-shaped). What is *not* reusable is the clock: a selection needs no timer, only the same press-to-release bracket.

**The drawing.** A selected range is a style over cells already drawn, which is the cheapest thing this renderer does. It costs no rect and no row, so §11.1's rule against a transient thing moving content is untouched.

**The copy, which is the hard half.** Two routes:

- **OSC 52**, the terminal-native clipboard escape. No dependency, works over SSH and through tmux, and is what a TUI should use. But it **cannot go through `Painter::put`**: [#120](https://github.com/breferrari/vigia/issues/120) established that `ratatui` filters control characters out of every string write, so this needs a direct write to the terminal handle outside the buffer, which is a seam this shell does not have yet. It is also **disabled by default in several terminals** (xterm wants `allowWindowOps`, others ship it off for the obvious reason that a remote process writing your clipboard is a capability), and there is no reply to read, so the program cannot know whether it worked.
- **A clipboard crate**, which adds a dependency `SPEC.md` does not name (so a spec change first), pulls X11/Wayland/AppKit/Win32 backends, and does nothing over SSH, which is a real way this tool is used.

## This is B9 reopening, and on the condition B9 named

`SPEC.md` §11.2 **B9** declined a yank key on 2026-08-15, and its stated objection is the one still standing: it would be the first key on this map that **destroys something the reader owns** (whatever was on the clipboard) with no way for the program to confirm it happened. OSC 52's silence is exactly that.

What B9 also recorded is the case *for* it, and this request is that case arriving from a reader rather than from an audit: the takeover captures the mouse, so the terminal's own selection is gone unless you know the modifier, and at the forty columns I6 is named for the path is elided, so there is no route to the real text at all.

So the ruling to make is narrower than "should there be copying": **given that a reader has asked for it, does an unconfirmable clipboard write become acceptable, and is the modifier bypass enough on its own?** Those are separable and the first is cheap.

## Gates

- The modifier bypass is documented and correct for the terminals named, checked by hand on at least one of them rather than cited from here.
- If in-app selection lands: a drag selects the cells it crosses and nothing else, a release keeps the selection, and any key clears it.
- Nothing about selection moves a rect, grows the footer, or changes a region's height (§11.1).
- Selection state is view state, not frame state: a write while text is selected redraws underneath it.
- If OSC 52 lands: the escape is asserted byte for byte the way `terminal.rs` already asserts the mouse bundle, and the fact that success cannot be confirmed is written into the spec rather than left for a reader to discover.

breferrari:
Closes #177. The build is #380 and lands in this pass.

**Nothing a reader can see has changed yet.** This PR is the ruling; #380 is what puts a wash on the screen.

## What is ruled

`SPEC.md` §11.2 **B20**, which was declined on 2026-08-29 and withdrawn by the reader the same day, is ruled the other way: **the shell draws its own selection over the diff, and the copy is resolved against the row model rather than read out of the cells.**

A left drag washes the rows it crosses. `y` sends those rows' real text: the whole line where the pane clipped it, one whole line where wrap split it over several rows, the repository-relative path where the row is a file heading, the `@@` line where it is a hunk header. With nothing selected, `y` is unchanged and sends the caret file's path.

## Why the decline was wrong, and it was not wrong about the cells

B20 held that a cell carries what was **drawn**, so a cell selection returns the elided label and cannot answer the complaint that produced it. That is true of the cells. It was never true of the program: `view::Row::Line` holds the whole line the core produced and `Row::Wrap` the whole tail, both independent of what the painter clipped. **Drawing a selection over painted cells does not oblige the copy to come from those cells.**

So the span is rows and the copy is the row model's, which reaches the three things the terminal's own bypass modifier cannot reach on any terminal at all. Half a line is what the bypass does well, and it goes on doing it.

## Evidence that did not fit the ceiling, recorded here instead

B20 set its own reopen condition as *"a terminal in real use that has no bypass"*. That condition has fired. Browser terminals built on xterm.js, which is what a reader inside code-server or a VS Code web session is typing into, are reported not to honour the Shift bypass, so there the pane holds the mouse and nothing hands it back. It is a report from elsewhere rather than vendor documentation, and it is the weaker half of the case: the stronger half is that the bypass could never reach a clipped tail on any terminal.

It is not in `SPEC.md` because the ruling had to be paid for in prose and the argument that survives without it is the one that belongs in a contract.

## Three sentences stopped being true and are corrected rather than left

- §11.1's click paragraph ended *"A click on the diff below stays inert, because nothing there is selectable."* It is where a selection begins now.
- §11.1's hover paragraph claimed *"no selection exists, nothing is remembered that a key acts on"*. Scoped to the mark that is still true; read as a general claim it is not, and the gate under it is re-described by what it actually holds, which is that motion alone is no act.
- §11.2 B9 read *"It is decidable only in the keymap, because an OSC 52 write draws nothing: the rendering suite stays green whichever way this goes."* A selection draws, so the rendering suite is no longer neutral on this.

## B4 does not reach it

B4's ground is that selection implies focus, focus implies a mode, and modes are reviewer-class. A selection the **pointer** addresses gives no region the keys' attention: no region becomes the one keys are talking to, and no key changes which region it means. Two keys change what they do and both resolve on ladders they already have, `Esc` leaving the frontmost thing and `y` sending the thing that can be named. Every other key does exactly what it did.

## The written layer was at all four ceilings, so this is paid for rather than added

Every document sat exactly on its budget, so the ruling had to buy its own bytes. What came out is prose the document restates elsewhere or that no longer holds:

- **Two vendor surveys** of how other tools wrap and select. §11.2 B10 already rules that a claim about other people's software lives in `RULINGS.md` and not here, and `README.md` carries the modifier list a reader actually needs. The wrap survey's conclusion is restated word for word by §11.1's own `w` paragraph.
- **The record of the cap** that was overruled and deleted, where the spec is present tense.
- **The paragraph arguing that nothing reading the pointer remembers it**, which B20 supersedes by drawing the line at focus rather than at memory.

`SPEC.md` falls 387648 to 387641 and `ROADMAP.md` 95476 to 95473, and both ceilings ratchet down with the total.

## Scope of checks

Documentation, plus three constants in `crates/vigia/tests/package.rs`. Only `package.rs` and `vigia-core/tests/coverage.rs` read these documents at runtime; the other twenty-six files that name `SPEC.md` do so in comments.

`cargo test -p vigia --test package --test register` and `cargo test -p vigia-core --test coverage`: **31 passed, 0 failed**, `the_written_layer_stays_under_its_budget` and `no_prose_paragraph_is_hard_wrapped` among them. No frame path is touched, so no budget gate runs.

## Plan fidelity

Two deviations from the plan on #177, both taken and justified here rather than at audit time:

- **No `RULINGS.md` entry for the cell-exact branch.** `RULINGS.md` sits exactly on its own byte ceiling, so adding the trail there means deleting unrelated prose from a ledger of past reasoning to make room. The branch compresses to one falsifiable sentence, which is what the house rule asks a ruling for, so it is in `SPEC.md` instead and the ledger is left alone.
- **`docs/THEME.md` and the new theme token are not here.** They are the build's, not the ruling's, and they move with #380.


breferrari:
## Review round: `/simplify`, four agents, four findings, all fixed

Ran read-only with the gate results in the brief, per the concurrency rule.

- **Stale sentence the first pass missed.** §11.1's hover paragraph still read "**Nothing in the diff body**, which is not clickable and must not imply it is". Found independently by the reuse and altitude agents, and it is the same claim the first pass caught one paragraph above at §11.1's click rule and missed here. Deleted; net negative on its own.
- **The B4 defence was a carve-out, and this is the finding worth the review.** It argued that a selection the pointer addresses acquires no focus. That is the loose reading of B4; everywhere else the document cites B4 it applies the strict one, *nothing may be remembered that changes what a key means*, and `y` fails that as written. Rewritten to answer the strict test: `y` means what it always meant, and what is nameable has been a fact about the screen since B9, where scrolling moves the caret and moves what `y` sends with it. A wash is a second thing the screen can name, not a second thing a key can be.
- **A deletion lost evidence with no other home.** The wrap-by-default survey named `lazygit`'s `wrapLinesInStagingView: true`, which the agent confirmed appears in no other tracked file. It was cut claiming §11.1 restates it, and §11.1 restates only the conclusion. The survey moves house instead: `RULINGS.md`'s B19 section, which already carries the tool-by-tool read, is where §11.2 B10 rules such a claim belongs.
- **A second deletion lost a verification.** "Checked by hand on Windows Terminal" is back in `README.md`, beside the modifier list it verifies.

Plus the stray third blank line the first deletion left, which the simplification agent located exactly.

**The ceilings trade rather than rise.** `SPEC.md` 387648 to 387356, `RULINGS.md` 94399 to 94676, written layer 581916 to 581898: the contract sheds more than the ledger takes, which is the trade `WRITTEN_LAYER_TOTAL` exists to permit. B20 was trimmed to pay for the above out of its own prose rather than out of the document's.

**Not taken, and why.** The reuse agent proposed shrinking B20's restatement of the row-model fact by pointing at §11.1; that is done, but only in part, because the sentence also has to refute the decline's premise and a bare pointer does not. The efficiency agent found no cost objection and asked for one wording clarification, that "re-resolved against every frame" not be read as rebuilding the sendable text every frame. The build does resolve both together on purpose, which is what makes the wash and the payload the same thing, so the sentence stands as written and the build's own comment says so.

Gates after the fixes: `package` 18, `register` 5, `sheet` 51, `coverage` 8. **82 passed, 0 failed.**

PR #382, opened 2026-09-02
A drag over the diff selects rows, and `y` sends their real text

breferrari:
Builds #380, the build half of #177. Stacked on #381, which carries the ruling: until that merges this branch's diff includes it.

**Draft, and honestly incomplete.** Pushed so the work is durable, not because it is ready. What is left is named at the bottom.

## What a reader gets

A left drag over the diff body washes every row it crosses. `y` then sends those rows' text resolved against the row model: the whole line where the pane clipped it, one whole line where wrap split it over several rows, the repository-relative path where the row is a file heading, the `@@` line where it is a hunk header. With nothing selected, `y` is unchanged and sends the caret file's path.

## Shape

- **`Selection` is two screen rows**, and `selection_after` moves them, sitting beside `hover_after` because a selection is pointer state rather than an intention. No `Action` is added.
- **The span lives on `Shell`**, so `Shell::apply` is the one place every action but `Yank` clears the wash, and where `Esc` gains its rung under the sheet and over quitting. A list of keys that clear would be a list somebody has to keep up to date.
- **The text is resolved in `App::view`**, beside the caret's path and for the same reason: what `y` would send comes from the frame that was drawn, never carried from the gesture, so the wash and the payload cannot disagree. `selected_text` is total over `Row`, and a wrapped line resolves to the row it continues, so it is emitted once however many rows it spans.
- **The wash is `theme.selection`**, painted after each row at the width the row's own wash uses so it meets the scrollbar the same way. `set_style` patches rather than replaces, so the syntax ink survives and only the background changes. On `ansi` it is `REVERSED` rather than a colour, for the reason the row washes there are unset: a background has to assume one and that is the palette that cannot.
- **No new wake and no timer.** `EnableMouseCapture` has written `?1002h` and `?1003h` since the takeover, so a drag already arrives at a wake this program pays for. Nothing here touches I1's soak.

## State of the gates

`cargo test -p vigia` is green except **`tests/sheet.rs`, 14 failures**, all one cause: a ninth mouse gesture moves the roomy rung from thirty-seven rows to thirty-eight and shifts the page counts with it, and the fixtures encode both. Four of the coupling gates that failed first are already fixed and passing, including `every_gesture_the_readme_teaches_is_named_on_the_sheet` and `no_gesture_token_hides_inside_another`.

Two things worth recording from that, because they are not obvious:

- **The `y` verb's width sets the sheet's rung.** Rewording it to "copy the selection or path" widened the keyboard group past what 80 columns can hold and dropped the whole sheet to its tight spelling, taking `PgUp` off it. It is "copy rows or path" now, inside the width the old verb already spent.
- **`SECTIONS` needed no edit**, because the roomy rung names the mouse group as a whole rather than by row.

## Left to do

- The sheet's geometry fixtures, which is the 14 above.
- `docs/THEME.md`'s row for `selection`, which `tests/theme_docs.rs` holds in both directions.
- A contrast gate for the new token in `dark`, `light` and `ansi`.
- `tests/select.rs`, whose load-bearing case is that `y` over a clipped line sends the tail that is not on screen.
- A screenshot, because Phase 8's bar is vetoed by eye.


breferrari:
## Finished. 1105 tests pass across both crates, 0 failures.

Writing the gates found a defect that would have shipped green, which is the point of writing them.

### A wrapped line was being sent truncated

`View::collect` cuts a wrapped line at its break points into **disjoint** slices: a `Row::Line` carrying only as far as the first break, and a `Row::Wrap` for each slice after it. The resolver emitted the head row and stopped, so with `w` on a reader got the line up to the wrap and nothing after.

The premise came from the row model's own docblock, which reads *"the whole tail"*. That describes what the painter is handed, not what the model holds. `line_at` joins the pieces back and `head_of` finds the row a continuation belongs to, including where its head has scrolled off the top of the pane. `tests/select.rs::a_wrapped_line_is_sent_once_and_whole` is what failed.

A second one from the same file: **a drag on the blank below a short diff selected the diff's last row**, because the span was clamped to the collected rows at both ends. A span starting past the end now selects nothing and `y` falls back to the caret path.

### What it looks like

Three rows crossed by a drag, marked here with `▐` because a background does not survive a paste. The third line is clipped on screen at `›` and arrives whole in the copy, which is the entire ruling in one picture.

```
   fixture · 1 changed                                                 watching

 ▸ A src/engine/watch.rs                  ■■■■■■■■■■■■ ____________    +5    -0
   A src/engine/watch.rs                  ■■■■■■■■■■■■ ____________    +5    -0
   @@ -0,0 +1,5 @@
▐  1 + fn advance(&mut self) -> Result<()> {
▐  2 +     let coalesced = self.settle(BURST)?;
▐  3 +     let very_long = one_of_those_lines_that_runs_well_past_any_eighty_c›
   4 +     self.emit(coalesced)
   5 + }
```

```
y →  fn advance(&mut self) -> Result<()> {
         let coalesced = self.settle(BURST)?;
         let very_long = one_of_those_lines_that_runs_well_past_any_eighty_column_pane(alpha, beta, gamma);
```

**This is a render, not a screenshot.** The feel still wants a human eye on a real terminal, which is Phase 8's own bar.

### The sheet costs a row, and the ladder pays

The roomy rung goes from thirty-seven rows to thirty-eight and the page counts move with it. `SPEC.md` §11.1 drops the two figures it restated beside its own sentence saying they are held by gates *"because a number here goes stale on the next gesture"*. This is the next gesture.

Worth naming because nothing documents it: **the `y` verb's width sets the sheet's rung.** Rewording it to "copy the selection or path" widened the keyboard group past what eighty columns hold and dropped the whole sheet to its tight spelling, taking `PgUp` off it. It is "copy rows or path" now, inside the width the old verb already spent.

### The contrast gate is relative, and my first draft of it was wrong

It held the wash to `TRACK_FLOOR` and to an invented 4:1 for ink, and failed. A background under a whole row and a one-cell mark on the pane are not the same problem: the shipped row washes sit at **1.27:1 to 1.57:1** against their panes and take `comment` to **2.62:1**, so a fixed floor drawn from a mark fails the elements it was modelled on.

The gate now asks that the selection be no worse than the diff wash it stands in for, on visibility and on every dim ink, which also keeps it honest as those washes move. The colour is the washes' own luminance in a hue neither uses.

### On timing gates, since it came up

`a_frame_under_the_roomy_sheet_holds_the_frame_budget` did fail, and **not on time**: its own shape guard caught that a 120x40 pane had stopped drawing the roomy rung, so the gate would have quietly timed a different sheet. That guard is the reason this change could not silently weaken a budget. Nothing here moved a frame-path number; the wash is a `set_style` over cells already drawn.

### Gates added

Twelve in `tests/select.rs`, one in `tests/palette.rs`. The load-bearing one is `the_sent_text_is_the_whole_line_where_the_pane_clipped_it`, which first asserts the pane does **not** draw the whole line, so it cannot pass for the wrong reason.

The escaping-file count moves to twenty-three in the four places that state it: `SPEC.md` §9, `Cargo.toml`, `package.rs`, `RELEASE-SMOKE.md`.

breferrari:
## Harden audit

Reference PR: **#375**, the build behind §11.2 B9. Same key, same feature family, +745 lines, and its own audit found six real bugs across five rounds. Two of them named the classes this change was exposed to and both recurred here in new clothes: a copy going stale across a batch, and an exit path that sent the wrong thing.

### Gates before the loop

- **`/simplify` (phase 4)**: four remits, **14 findings**. Themes: three real defects, plus reuse of `hover_repainted`, `render::span` and `Regions`' sheet test.
- **Plan fidelity (phase 5)**: run by `take-next` before the gates were written, and **it was wrong to carry that result forward**. #380's exit criteria promised a gate that a hunk row sends its `@@` line and `select.rs` contained no `@@` at all. Re-diffed after the gates existed; that omission is what let the hunk-header defect through.

### Rounds

| Round | Remits | Findings | Real defects |
|---|---|---|---|
| 1 | reuse, simplification, efficiency, altitude | 14 | 5 |
| 2 | adversarial, gate-and-ratchet, spec | 16 | 3 |
| 3 | adversarial, correctness, spec | 13 | 3 |
| 4 | adversarial, gate-and-ratchet | 4 | 0 |
| 5 | fresh-eyes full re-read | 4 | 0 |

**Trajectory: 14, 16, 13, 4, 4.** Rounds 1 and 2 are not comparable, since they ran different remits; from round 2 the curve is strictly decreasing on matched remits. **Exit: dry on behaviour at round 5**, with the last two rounds finding nothing a reader could observe. Self-introduced fraction stayed low: one finding in round 2 lived inside round 1's fix, and one in round 5 inside round 4's.

### Round closers

- **Mutation (decider and producer)**: nine mutations applied, each reverted by inverse edit with the file hashed before and after, because the tree was dirty. **Eight killed, one survived** — and the survivor is the most useful result in this audit: `a_wrapped_line_at_the_foot_of_the_pane_is_still_sent_whole` passed with the fix deleted, because its fixture drew the whole line. Rewritten to assert the loss before the recovery, it now reddens. Killed: the `View::whole` lookup (twice, before and after the dead join came out), the wash-clearing in `App::select`, the empty-payload filter, the `Theme::resolve` depth fallback, `chrome.selected = repainted(...)`, `holds_a_selection` reading the raw span, the singular branch in `Yanked::lines`, and the empty-pane retire in `Shell::paint`.
- **Coverage shape**: the iterated surface here is the diff row loop, which the wash rides. Checked that `Shell::paint` runs twice under `owes_repaint()` and that the second pass sees `was` as the first pass's own `painted`, so nothing is retired twice.
- **Rules bite**: every gate this PR adds has a named one-line violation that reddens it. Three could not fail when written and were fixed: the order gate over `Shell::draw` grepped for the hover mark alone; the heading gate drew its path whole at eighty columns; the wrapped-line gate passed with wrapping off. Each now asserts its own precondition first.
- **Measurement instruments**: no perf gate was added. The one budget failure during the work was `a_frame_under_the_roomy_sheet_holds_the_frame_budget` failing its **shape guard**, not its time: a 120x40 pane had stopped drawing the roomy rung once the sheet grew a gesture, so the gate would have quietly timed a different sheet. Pointed at 120x41, which makes it time a taller sheet than before.

### Real defects found

1. **A wrapped line at the foot of the pane was sent truncated at a column** (`view.rs`), which is the one thing B20 exists to prevent. The lossy window is one fixture-copy wide: fewer pieces and the line fits, more and `taller_than_pane` extends the last piece and loses nothing.
2. **An empty selection wiped the reader's clipboard** (`app.rs`), while the footer said it had sent a line. §11.2 B9's surviving ground is that the clipboard is state this program does not own.
3. **A click on a clean worktree took the quit key** (`lib.rs`). A span over rows the region owns and the walk had nothing for drew nothing, resolved to nothing, and still answered the `Esc` rung.
4. **The wash was invisible below truecolour** (`theme.rs`). `Depth::resolve` drops every background and, unlike the diff washes, the selection had no bar to degrade onto.
5. **A cleared wash still yanked** (`lib.rs`, `app.rs`), at the new region's rows after a tick that moved it. Same root as `Esc` then `y` in one batch sending the cancelled selection.
6. **The hunk header was spelled a second way** (`view.rs`). `render::span` drops the `,1` git drops, and a header spelled the other way names a different line.
7. **`Esc` stopped leaving the frontmost thing** (`input.rs`): a press beside the sheet was selectable, so a wash could stand under an open sheet and take `Esc` from it.
8. **Focus loss discarded the wash** (`input.rs`), on a tool built to sit beside a pane a reader types into.
9. **The span was anchored to the previous paint's geometry** (`lib.rs`), so the wash and the payload could name different rows.
10. **A failed collect kept stale lines** (`lib.rs`).
11. **A drag on the blank below a short diff selected the last row** (`view.rs`).
12. **`line_at`'s tail join could not run** (`view.rs`), unreachable for any `View` this crate builds.

Plus five spec contradictions: the hover rule's criterion picking out a fourth surface, the sheet mockups two gestures behind a drawer they claim to be rendered from, a counter reading `of 22` against a table of 25, a forward pointer promising a scroll in the branch that ships a clamp, and `y` described as sending "the selection where one stands" where it falls back when it names no line.

### Tests added

**Twenty new gates**: nineteen in `crates/vigia/tests/select.rs`, two in `crates/vigia/tests/palette.rs`, one source-form gate in `crates/vigia/src/lib.rs`. Categories: unit (the pure `selection_after`, `Selection::rows`, `Region::clamped_row`), integration (the resolver end to end through `App::view` and `Action::Yank`), rendering (the wash's extent and that it moves no rect), and contract (contrast against the wash it stands in for, and survival at every colour depth).

**Tests before to after: 1094 to 1114**, 0 failures, across `vigia` and `vigia-core`.

### Deferrals

None. Every finding was fixed in this PR.

### One recorded decision

`y` emits **no trailing newline** on a multi-line selection. The field is split, so this is a choice rather than an oversight: [tmux excluded it through 1.9a and began including it in 2.0](https://github.com/tmux/tmux/issues/61), contentiously. With no selection `y` sends a path, which must not carry one, and a key that terminates its output in one mode and not the other is worse than either rule.

Issue #416, opened 2026-09-04
The line number becomes the note icon under the pointer, and the note rows draw under the line


breferrari:
Child of #414, parts 2 and 3, and its mockups. The mark-only half: everything the pane draws for a note, with the note's text coming from the store rather than from a box.

## Build

The pointer resting on a content row's gutter (number, sigil, gap) turns the number into the icon in `bar_hover`'s ink, through B10's mechanism and its clearing ladder. A click there writes a note with an empty body to the store, which is the anchor alone, and draws the icon durably on that line. A click on the icon of a line that already carries a note withdraws it: the note departs and its file is deleted, the reader's own words leaving on the reader's gesture. Every note in the store that resolves to a drawn row is drawn under it as display rows: a bar down the left in the note's ink, the body in the chrome's dim weight wrapped at the content width, the status word at the right edge, the agent's reply under it when there is one. A note whose line was edited draws under that number in a dimmer weight with the word **changed**; a note whose line is gone draws under its file's heading with the word **gone**; a note whose file is not in the diff draws nowhere and is counted in the footer as `2 notes · 1 adrift`. One open note per line: a click on a noted line withdraws it here, and reopens the box once the box exists. `c` hides and shows the note rows. The rows are display rows the bar does not count, inserted where `wrap_rows` inserts continuations. `✎`, `▎` and `↳` take their rungs on §5.1's CP437 ladder.

## Gates

The hover mark appears on the gutter and on no content cell, and clears by B10's ladder. A click on the gutter writes one file and a press on content still begins a selection. Note rows are drawn under the anchored row at 80 and at 40 columns and never over-occupy, inside `legibility.rs`'s sweep. A gate per row of #414's edge-case table that the pane owns: moved, changed, gone, adrift, renamed, deleted, two notes on one line, an unwritable store keeping the text and saying so in the footer, a torn file skipped with one notice. The thumb's height does not move when a note's rows are drawn, which is the seam #412 and #413 fix first. A frame with fifty notes on screen holds the 16ms budget, measured in `tests/budgets.rs` with and without the notes, interleaved.

## Blocked on

#413 and #412, and the store from the first child. The mode is not in this issue: a click here writes an anchor, and the box is its own child.

breferrari:
## Plan, before implementing

[...]

Issue #414, opened 2026-09-04
decision: a note on a diff line reaches the agent over MCP


breferrari:
Asked by the reader on 2026-09-04, from use. This is a decision issue: it proposes §11.2 B21, shows what it looks like, names what it costs, and asks for rulings before its children are built. The children are filed in the same phase and blocked on the rulings they need. Amended the same day by the reader: what happens once the agent has read a note, which is a departure rather than a state, and the box's own entrance.

## What is asked

Hover the left side of a diff line and an icon appears. Click it and a proper box opens under the line, pushing the diff below it down, with an input where the reader types a note for the coding agent in the other pane. The box is a box: bordered, its own ink, arriving through an effect, so a note can never be mistaken for a line of code. The agent reads the note over MCP, anchored to the file and line, and can answer it. The reader runs several `vigia` panes on several projects at once, so a store or a server that assumes one instance is out. Two crates were offered as prior art, `ratatui-textarea` and the `tui-widgets` workspace, and a question was asked about Claude Code's message system carrying a send button straight into the running session.

## What the record says, and what binds

`ROADMAP.md` lists under permanent non-goals: *"Annotations and comment threads. Reviewer-class by definition."* `SPEC.md` §4 says the same: *"annotations, comment threading ... Each is reviewer-class and each would cost an invariant."* Neither line carries `Ruled <date>, reader`, so by `CLAUDE.md`'s rule both are session-inferred and do not bind the reader. If this is ruled in, both lines go in the same commit, not annotated.

What does bind is the reasoning underneath, and each piece is answered below rather than waved at: B4, *selection implies focus, focus implies a mode, and modes are reviewer-class*; §11.1's *the monitor writes nothing, not into the worktree, not into a state directory, not into `.git`*, on which B7 and the settle margin rest; I1, I4 and I9; and the non-goal *AI features of any kind*. §2's table is the test every part has to pass: correct when untouched, runtime days, latency a frame.

## The frame that fits the thesis

`btop` carries a signal menu. A monitor may hold a bounded act on the thing it watches, and that does not make it a reviewer. Here the thing on the far side of the diff is the agent, and the act is one sentence to it, pinned to a line. A pane with no notes draws what it draws today, so *correct with zero interaction* is untouched. Nothing here calls a model: the pane is a mailbox and the agent reads it, so the non-goal about `vigia` judging files stands. The one real cost is a mode, and it is named in part 5 for the reader to rule on.

## How it looks

[...]

Issue #178, opened 2026-08-15
The absolute frame budgets flake on shared runners, and the failure reads as a regression


breferrari:
Found while merging [#166](https://github.com/breferrari/vigia/issues/166): the absolute frame budgets fail on shared CI runners at a rate that is now visible in the record, and each failure looks exactly like a regression.

## The two occurrences, one day apart on different tests and different platforms

- `main` at `34f74ec`, windows: `the_frame_budget_holds_through_a_bulk_rewrite` reported **p99 98.96ms** against the 48ms budget, with **p50 9.25ms and max 255.40ms**.
- `step-buttons`, macos: `a_frame_holds_the_budget_however_deep_the_reader_has_scrolled` reported **p99 73.14ms** against the same budget, with **p50 3.67ms and max 179.78ms**. It passed on a re-run of the identical commit, and passed on ubuntu and windows in the same matrix.

## The shape is the finding

In both cases the **median is comfortably inside budget and the tail is an order of magnitude out**. A real regression moves the median; a runner losing the CPU for a scheduling quantum moves two samples out of 250 and nothing else. So the number these gates report on failure is not evidence about the code, and it reads as though it is: the assertion message quotes a p99 against a budget and says "over", which is the sentence a reviewer acts on.

`SAMPLED_FRAMES` is 250, so this is not the *"p99 over 30 samples is the maximum"* trap. The percentile is honest about the distribution it was given; the distribution is what a shared runner produced.

## Why it is worth fixing rather than re-running

A gate that fails for reasons unrelated to the diff gets re-run until it passes, and a gate that gets re-run until it passes has stopped being a gate. The [#142](https://github.com/breferrari/vigia/pull/142) PR already recorded one of these and reported it honestly rather than dropping it, which is the right handling and also the point: the same handling twice is a pattern, and the third time nobody will read the numbers.

## Candidate directions, none of them ruled

- **Report the shape on failure, not just the percentile.** The assertion already prints p50 and max. Making the *failure text itself* name the diagnosis when p50 is inside budget and only the tail is out would stop a reader having to know this issue exists.
- **Discard a bounded number of worst samples**, which is what a trimmed percentile is for, and state the trim in the message. This weakens the gate by exactly the amount stated, rather than by an amount nobody wrote down.
- **Fail only on a repeat**, running the sample loop twice and requiring both to breach. Doubles the wall clock of the slowest job.
[...]

PR #436, opened 2026-09-06
A bordered box under the line takes the note

### LINKED ISSUE #419: A bordered box under the line takes the note
Child of #414, parts 4 and 5, and its mockups. This is the one that carries the mode, and it is not taken until #414 rules A.

## Build

A click on the gutter opens a bordered box under the line at the content's origin, the anchor in its top edge and `Enter sends · Esc cancels` in its bottom edge, drawn in the chrome's ink so nothing in it reads as code. The editor is `ratatui-textarea`, `default-features = false` with the `crossterm` feature, named in `SPEC.md` §6 with what it adds to the graph (one crate) before it reaches the manifest. The box grows to a cap of a few rows and scrolls inside it. Enter writes the note and closes the box, Esc closes it and writes nothing, an empty body writes nothing, a click outside it closes it and writes nothing, the wheel still scrolls the region behind it. While it is open, keys go to it and nothing outside it changes meaning. It arrives through `tachyonfx` with an entrance worth watching: the border draws itself in from the anchor's corner and the body fades up behind it, over the same duration a changed file takes; Esc plays it backwards and Enter hands off to the note rows' own arrival. Armed by the click, done when done, never a clock of its own. Alt+Enter inserts a newline where the terminal reports it; v1 is one paragraph.

## Gates

Every key the map names does nothing to the pane while the box is open, and everything again once it closes. Enter writes exactly one file and Esc writes none. The box at 80 and at 40 columns inside `legibility.rs`'s sweep: never over-occupies, the anchor label degrades the way every label does. A frame with the box open and its entrance running holds the budget, and the entrance ends inside its duration with no wake of its own. The dependency is named in §6 and gated by #403's check once that lands.

## Blocked on

#414's ruling A, and the second child's drawing.

breferrari:
Closes #419. The build half of `SPEC.md` §11.2 B21: a press on a content row's gutter opens a bordered box under the line, the reader types the note there, and Enter writes it. Taken with `take-next` on 2026-09-06, unattended, with plan approval waived at dispatch; the plan is on the issue.

## What a reader can now do

Click a line's number in the diff and a box opens under it with `note · src/watch.rs:5` in its top edge and `Enter sends · Esc cancels` in its bottom. Type. Enter writes the note and its rows arrive under the line; Esc closes it and the box leaves the way it came. Click a line that already carries a note and the box opens with its text: Enter rewrites it, an emptied box withdraws it. Every key is the box's while it is open, a click anywhere else closes it, the wheel still scrolls behind it, and the agent's next write does not move the viewport out from under it.

## Scope of the checks

The diff is Rust across the frame path, the walk, the painter and the loop, plus the manifest and the documents. The full suite, clippy under `RUSTFLAGS=-D warnings`, `cargo fmt --check` and the release budget gates all ran. Debug: **1337 passing across 58 binaries** with fail-fast off, zero failures. Release, alone on the machine: **23 of 23 budget gates green**. The box's own gate reads **p50 12.49ms** against I9's 16ms, with a first round at p99 17.89ms and a second at 14.19ms, which is #178's re-measure rule doing its job; the same fifty notes without the box read 12.54ms, so the box costs nothing this instrument can see.

## The dependency's cost, measured again

B21 recorded `ratatui-textarea` as one crate, which was the lock file's answer. Over the build graph it is **six crates on Windows and seven on musl**: itself, and `time` with `time-core`, `deranged`, `num-conv`, `powerfmt` and, on unix, `num_threads`. The crate declares `ratatui-widgets` with that crate's defaults, `calendar` among them, and feature unification is additive, so the `time` tree the shell's own `ratatui` pin had kept out by taking no widgets arrives through the editor and nothing on this side can send it back. All pure Rust, no `build.rs`, and the `no C toolchain` job holds that. §6 carries the measured number, `RULINGS.md` B21 the correction beside the count it supersedes, and the `ratatui` pin's comment no longer claims `time` is out. The ruling named the editor and the build takes it; the number it named was true of the lock file and false of the binary, and that is recorded rather than quietly corrected.

## Plan fidelity

Every promise in the plan is delivered, with these named differences, each taken deliberately:

- `BoxRoute` gained a `Paste` variant the plan did not list, so bracketed paste goes into the box rather than being dropped.
- `Esc` and a press outside the box were planned as two routes and are one, since both close the box and do nothing else.
- The forty-column gate is `at_forty_columns_the_box_takes_the_content_width_and_the_label_drops_its_head`: with this fixture's path the whole anchor fits at forty, so the ladder's rungs are asserted at twenty columns and across the legibility sweep, which sweeps every width from 1 to 120 with a longer path.
- `lib.rs`'s gutter-press source gate was rewritten for the box's shape rather than only extended.
- `SPEC.md` was funded by cutting trail prose the ledger already holds.

## Decisions taken without asking

Reopen on a noted line rather than withdraw in one click, because B21's third paragraph rules it and #414's body carries both readings. The caret is the cell reversed, the editor's own default, which adds no glyph to §10's residue. Border and hints in `chrome_dim`, the reader's text in `chrome`. Follow holds the viewport still under an open box; the reader's own wheel does not. The body cap is four rows. A rewrite returns the note to *open*. A note that resolved or vanished while its box was open is written as a new note on Enter. Ctrl-C inside the box goes to the editor like every other key, and Esc is the way out. On a pane too narrow to draw the box, and on one narrowed under an open box, the box does not exist: a mode the reader cannot see is one they cannot leave on purpose.

## Harden audit

Reference PR: #435, the previous B21 build on this surface. Its gates: a failing gate per behaviour, an interleaved with-and-without budget gate, a mutation set over the decider and the producer, the spec amendment funded under the ceiling, the plan in the PR body, a Sonnet simplify with documentation in scope, capped per-remit audit rounds, and test counts before and after. No property-test crate exists in this repository, so no property tests.

### Gates before the loop

**`/simplify` (phase 4).** Four Sonnet agents on reuse, simplification, efficiency and altitude. Nine findings, seven distinct, six fixed in 93feec5: `Timed` is the one retirement rule the box's effect and every note effect share, where two byte-identical copies stood; `effects_running` is one predicate where the frame clock and the record of whether an effect drew asked the same four-term question in two places; the note the box covers is left out where a file's notes are gathered rather than filtered at each of two readers, which deleted `BoxPin::over` and `Walked::covers`; `content_cells` is the arithmetic `box_cells` and `note_cells` had each spelled out; `Esc` and a press outside became one route; `box_edge` takes its corners as a pair, dropping an argument-count allow; and three `App` methods that asked one guard ask `open_mut`. Two skipped, both efficiency: the box's rows are built twice on a screen shorter than the pane, which is the note pins' own pre-existing pattern one line above, and the box's cells are computed for every event while only a press reads them. Both are a few hundred nanoseconds inside a frame measured at 12.49ms, and a cost this instrument cannot see is not one worth complexity.

**Plan fidelity (phase 5).** Run by `take-next` before harden and not repeated; its result is above.

### Rounds

Each round: Sonnet agents in parallel with disjoint remits, reports capped under 400 words and ranked worst first inside each remit, never reranked across them; fixes; then the orchestrator's own end-to-end re-read.

| Round | Remit | Findings | Worst in the remit | Fixed in |
|---|---|---|---|---|
| 1 | adversarial | 3 | a box opened on the last drawn row lost its body and both edges, so the reader typed into a top edge | 188f851 |
| 1 | correctness | 5 | the ring the border draws along had no unit test, and its degenerate one-row arm is reachable | 188f851 |
| 1 | spec steward | 1 | §6 claimed every crate the editor adds was already in the lock file, which is false of the editor itself | 188f851 |
| 1 | gates and ratchets | 2 | the written layer's total was raised 202 bytes on a ledger exception whose ruling is still open | 188f851 |
| 1 | the re-read | 1 | the source gate over `settle_notes` did not name the box's two steps | 580e2a1 |
| 2 | adversarial | 1 | a box taller than the pane kept its bottom edge and lost the caret | 580e2a1 |
| 2 | correctness | 1 | a press inside the box's exit opened an empty box over the note it held, and Enter wrote a second | 580e2a1 |
| 2 | TUI glance | 3 | a pane too narrow to draw the box still gave it every key | 580e2a1 |
| 2 | gates and ratchets | 3 | nothing drove the two clamps wanting different amounts at once | 580e2a1 |
| 3 | adversarial | 3 | the caret's floor ran after the bound at the line, so it could carry the box's top edge off too | 14d3d01 |
| 3 | correctness | 3 | the room rule was asked only at the press, so a resize under an open box left an invisible mode | 14d3d01 |
| 3 | gates and ratchets | 3 | the narrow pane's refusal was pinned by matching source text and never by a press | 14d3d01 |
| 4 | the closers and the re-read | 3 | the gate for the bound at the anchored line proved nothing, for the second time | 536e3d0 |

Trajectory: 12, 8, 9, 3.

**Exit: stopped at round 4 with nothing outstanding, and its two agents never ran.** Both died on a session rate limit before reporting. The round was finished without them, by the orchestrator's own re-read and by the mutation set, which is what round 4 was going to attack: six of its mutations target round 3's own changes. Two questions put to round 4 are settled by proof rather than by reading. `ceiling` is a minimum of two values each at least `floor`, so `clamp(floor, ceiling)` cannot panic. And the value handed to that clamp is at most `total - height`, while the clamp only raises it to `floor`, itself at most `total - height`, so the pane is never left short of rows it had.

This is a decreasing trajectory rather than a dry round, and it is named as one. What a fifth round would look at is under *Known and recorded*; none of it is unexamined.

### Round closers

**Mutation, decider and producer.** Twenty-six mutations across `commit`, `box_route`, `opening`, `has_room`, the walk's placement, the clamp's three terms, the caret's index and arithmetic, `box_cells`, the marks, `patience`'s new clock, `effects_running` and `Timed`'s retirement. **23 killed, 3 survived, none invalid.** Of the three: two were the set aiming at the wrong suite, since those gates are unit tests in the library rather than integration ones, and both die when run there. The third was real and is the round's own finding: the gate for the bound at the anchored line proved nothing, because its fixture was a twelve-line file on a ten-row pane where the whole diff fit and no bottom clamp existed. That fixture is a sixty-line file scrolled past its end now, and it asserts the window moved before it presses; removing the bound fails it.

**Rules bite.** Every guard this branch adds was watched going red against a deliberate violation before it was trusted: the clamp's caret floor, the bound at the anchored line, the leaving box's lookup, the narrow pane's refusal, the box's row cap, `box_cells`' extent and the departing press. One was watched failing twice, once for the missing rows and once for an off-by-one in the first fix. One mutation was found to be redundant rather than untested and its branch was deleted: `caret_in` had a case for a caret among the blanks a break dropped, and a break lands on a blank only where the piece already fills the row, so the branch below returns the same answer.

**Coverage shape.** The Rig in `tests/notes.rs` mirrors the loop's routing by hand: the press with both of its refusals, the key, the paste, Enter, Esc, the clock and the leaving box's lookup. It asserts it is never asked to press with a box open, which is the loop's own first branch. What it cannot mirror is the store with no home, since its store always exists; that guard is pinned by a source gate instead.

**Measurement instruments.** The release gate re-arms the entrance whenever it finishes and asserts it ran on all 300 timed frames, so no frame it timed drew the box still. It reads p50 12.49ms against 16ms, and the same fifty notes without the box read 12.54ms. No budget, ceiling or ratio ceiling moved to let this change pass: across the branch `ROADMAP.md` fell 570 bytes, `RULINGS.md` rose 516 and `SPEC.md` rose 54, so the ceilings net zero and the written layer's total is unchanged at 632780 with the same 48 bytes of slack `main` had. The comment ratio ceilings were hit four times and paid for by condensing comments each time.

### Real bugs found and fixed

- A box opened on the last drawn row lost its body and both edges.
- A box taller than the pane kept its bottom edge and lost the caret.
- The caret's floor, applied after the bound at the anchored line, could carry the box's own top edge off with it.
- A pane too narrow to draw the box took every key anyway, and a pane narrowed under an open box kept them.
- A press inside the box's 250ms exit opened an empty box over the note it held, and Enter wrote a second note on that line.
- §6 said every crate the editor adds was already in the lock file, which is false of the editor itself.
- The written layer's total was raised on an exception whose ruling is open, and is funded with prose instead.

### Known and recorded

- **A resolve that lands between `commit`'s read and its write is overwritten.** The read sits as close to the write as it can and the residual window is named in the comment, which is what the store's own rewrite does one layer down. No portable primitive closes it.
- **`box_over` matches the leaving box by exact anchor.** An edit to that line inside the 250ms exit makes the anchors differ, so a press in that window opens an empty box and Enter writes a second note on the line. Narrow, and its outcome is a state B21 already allows two panes to produce; a looser match risks reopening the wrong note, which is worse.
- **A box open while its file leaves the diff draws nowhere and keeps the keys.** Esc and Enter both still answer. B21 rules an adrift note is withdrawn through the agent with no footer gesture until use asks for one, so nothing is invented here.
- **At the `ansi` rung the pointer's hover and a noted line draw the same bold grey**, because that palette bakes the weight into `bar_hover` and `Mark::ink` adds it again. It shipped with #416; this build makes the anchor trace load-bearing rather than causing it.
- **A note another pane writes arrives with no effect.** `Ledger::reload` answers only what moved; the reader's own Enter arms its arrival locally.
- **`BOX_FLOOR = 7` is a regression pin**, read off the sweep rather than derived from the box's own arithmetic.


breferrari:
Thanks. That concern is the right one and the audit reached it too, from a different direction. It is three cases, and they do not have one answer.

**The pane is too narrow to draw the box.** Fixed, twice. `notes::has_room` refuses the press and the footer says so, and since a pane can also be narrowed *under* an open box, `Shell::settle_notes` asks the same question on every path to a paint and sends the box away the way Esc does. Gated by `a_pane_with_no_room_for_the_box_opens_none`, which presses a six-column pane whose gutter still answers, so the width is what refuses it rather than the absence of a target, and by a mutation of `has_room`'s own bound that goes red. Before that fix the pane read as an ordinary diff while Enter would write a note that had never been on screen, which is exactly what you describe.

**The box is scrolled off by the reader's own wheel.** Not a defect: `SPEC.md` §11.1's B21 paragraph rules that the wheel and a resize pass through to the pane while the box is open, so the diff moving behind it is the ruled behaviour. Closing the box there would contradict the ruling, and the reader who scrolled can scroll back.

**The box's file leaves the diff while it is open**, because the agent committed or reverted it. The box then draws nowhere and keeps the keys. This one is recorded rather than fixed, and the reasoning is in the PR body under *Known and recorded*: nothing is lost and nothing is stuck. Esc still cancels, Enter still writes, and the note it writes becomes an adrift note, which the footer already counts beside the position as `2 notes · 1 adrift`. B21 rules that an adrift note is withdrawn through the agent, with no footer gesture until a week of use asks for one, so a new element announcing this state is a feature nobody has asked for. If use does ask for it, the cheapest form is one sentence in the footer's existing alert voice, said once, and it needs the frame's file list threaded into `settle_notes`.

So: one fixed and gated, one ruled, one recorded with its reason.

PR #489, opened 2026-09-09
An mcs tech pack installs the notes server and its hooks

bguidolim:
The README asks a reader to register `vigia mcp` by hand and to paste three hooks into their own settings. This packages both as an [`mcs`](https://github.com/mcs-cli/mcs) tech pack, so a reader who keeps their Claude Code setup that way gets the same configuration installed and maintained instead, and `mcs doctor` tells them when it has drifted.

`mcs pack add breferrari/vigia` then `mcs sync --global`. Nothing about the binary or the pane changes, and a reader who does not use mcs sees one new line in the README pointing at `docs/TECHPACK.md`.

## What it installs

| Component | What |
|---|---|
| `vigia.vigia-cli` | `brew install breferrari/tap/vigia` |
| `vigia.notes-server` | `vigia mcp`, registered as `vigia` at user scope |
| `vigia.hook-register-start`, `-end` | `vigia mcp register` on `SessionStart` and `SessionEnd` |
| `vigia.hook-pending` | `vigia mcp pending` on `UserPromptSubmit` |
| `vigia.notes` | A `CLAUDE.local.md` section carrying the standard a resolved note is held to |

## Three choices worth checking

**The hooks are scripts, not the bare commands the README shows.** mcs only registers a hook whose install action copies a file, so a bare command has to go through a settings file instead, and those get presence-only verification: doctor can see that the event exists and never that it is yours. Going through a file earns a real check against the registration, makes each hook selectable, and lets the script exit clean when there is no `vigia` on the machine, where the bare command would put `command not found` in front of every prompt. It costs one bash process: measured interleaved over three rounds of 60, the wrapper runs at ~8ms against ~5ms for the command alone, so ~3ms on the prompt path.

**The binary installs through `shell:` rather than `brew:`.** `brew:` derives a `commandExists` check against the string it is handed, and `breferrari/tap/vigia` is not a command, so doctor would sit red on a working install. `shell:` derives nothing, which is why the check is written out: it is also what tells sync the install can be skipped, and without it every sync shells out to a tap round-trip. The tradeoff is that a shell action is never reversed, so `mcs pack remove` leaves the formula behind. That is the outcome to want here, since dropping the Claude Code integration should not take away the pane the reader is watching.

**`ignore:` is what stops this from being noisy.** The manifest sits at the repo root, so the crates, the specs, the docs and the assets all sit beside it and none of them are the pack. Listing them silences the unreferenced-file warnings and, more importantly, stops a commit that touched only `crates/` from telling every installed copy that a pack update is waiting.

## Open, and needs a ruling before it can be written down

`SPEC.md` section 9 enumerates the distribution surface: crates.io, the Homebrew tap, GitHub releases, no domain. This adds a fifth thing shipping from the root and the section does not name it, which is the kind of drift the method says to stop on rather than let sit. Two ways to settle it, and both are a decision rather than a cleanup: add a bullet naming the pack and what it distributes, which is Claude Code configuration and not the binary, or rule that the pack sits outside the spec's remit and say so once. `ROADMAP.md` carries no row for it either, for the same reason.

## Verification

`mcs pack validate .` passes clean against mcs 2026.9.3. Both hooks were exercised against the real binary and against a stub: stdout reaches the prompt, the stdin payload reaches `vigia mcp register` intact, a non-zero exit from the binary is swallowed, and with `vigia` off `PATH` both print nothing and exit 0. `sh .claude/scripts/selftest.sh` passes.

There is no Rust toolchain on the machine this was written on, so `register.rs` was not run. Its two markdown gates were replicated instead and both new documents pass: no hard-wrapped paragraphs, no malformed table rows, no em-dashes. The comment gates walk only `.rs` under `crates/`, which this touches none of. CI is the real check on that.

One invariant here has no test, and it should be said plainly: the `ignore:` list asserts that nothing outside `hooks/`, `templates/` and the manifest is pack material, and it is a snapshot. Add `deny.toml` or `benches/` later and validate grows a warning while every installed copy silently starts being told an update is waiting. That wants a gate in the shape `register.rs` already establishes, and it is not in this change.


breferrari:
Tracked as #505, on the Phase 7.5 milestone, since what the pack packages is the notes server and its hooks and Phase 7's filter is about putting the binary in someone's hands.

Two things are ours rather than yours and are being handled here: the `SPEC.md` §9 ruling you correctly refused to make, and approving a workflow run, since no CI has ever run on this branch. Your two markdown documents were checked by hand against the PR head in the meantime and both pass the gates `register.rs` would have run: no hard-wrapped paragraphs, no em-dashes, table rows well formed.

Two small things are yours, and they are the only changes being asked for. `docs/TECHPACK.md` never says the pack is macOS and Linux only, and it installs through `brew` with `#!/bin/bash` hooks while Windows is a tier-1 target for `vigia`, so a Windows reader following that page hits `brew: command not found`. And `hook-register-start` and `hook-register-end` both install `hooks/register.sh` to the same destination: what happens under `mcs sync --customize` when a reader selects one and not the other is something you will know and we cannot verify from here.

The `ignore:` gate you flagged yourself is recorded on #505 and is not being asked of you.

The reason this took a moment: writing the §9 bullet needed bytes the written layer did not have, every ceiling being equal to its file's size. #502 fixes that.

breferrari:
Pushed one commit to this branch (e6213de), moving the pack's own files under `.techpack/`. `hooks/` and `templates/` were two new top-level directories for one optional integration, beside `crates/`, `docs/`, `assets/` and `xtask/`; what the pack adds to the root is now `techpack.yaml` alone, which mcs discovers there and cannot move.

`docs/TECHPACK.md` stays put. It is a document a reader is sent to from the README, and `docs/` is where `THEME.md` already lives.

Worth recording because it is a trap: the first attempt at this move was `.claude/techpack/`, and it does not work. The schema forbids listing in `ignore:` any path a component references, `.claude/` is listed, so the manifest would have been invalid rather than merely untidy. The comment above `ignore:` now says the pack's subtree is absent from that list by requirement rather than by oversight. With `.techpack/` the list is unchanged.

Executable bits survived the rename (both hooks are still 100755) and the YAML still parses. I could not run `mcs pack validate .` because mcs is not on this machine, so that check is yours.

The two small things in my earlier comment are still the only changes being asked of you.

breferrari:
Your `isRequired` call was the right one and it is in, as 584edb0.

Both register components carry `isRequired: true`, and that also undid something I had done a commit earlier: I had given them separate destinations (`register-start.sh` and `register-end.sh`) to dodge the case where deselecting one removes a file the other needs, which the schema documents for cross-pack collisions and not for two components of the same pack. Required removes that case at the source, so the shared `register.sh` is back and the duplicate copy is gone. One mechanism instead of two.

`docs/TECHPACK.md` had to move with it. It said every hook was selectable under `--customize`, which is now false, so it says `hook-pending` is selectable and the register pair is not, with the reason: they are one mechanism split across two events, and a session that records its socket and never clears it leaves a stale one.

Everything else on this branch is ours and is described in the two commits above it: `SPEC.md` §9 now names the pack and rules that it distributes Claude Code configuration rather than the binary and is not a fourth way to get `vigia`; the pack's files moved to `.techpack/` so the root gains `techpack.yaml` alone; `docs/TECHPACK.md` says the route is macOS and Linux, since it installs through Homebrew and the hooks are shell scripts while `vigia` is tier-1 on Windows; and there is a `ROADMAP.md` row on Phase 7.5, tracked as #505.

Four paths changed under you and are worth a `mcs pack validate .` on your side, because mcs is not on this machine and that check is the one thing here nobody has run: three `source:` entries and one `contentFile:` now point at `.techpack/`.

`cargo test -p vigia --test package` is 32 passed and `--test register` is 6 passed, which is the run this branch had never had.

